PR #1, opened 2021-09-29
Adding FHIR $apply operation example.

dwwinters:


apply/screening-example.js
```diff
@@ -0,0 +1,118 @@
...
+const errorActivity = JSON.parse(readFileSync('fsh-tank/fsh-generated/resources/ActivityDefinition-CommunicateErrors.json'));
+const communicateActivity = JSON.parse(readFileSync('fsh-tank/fsh-generated/resources/ActivityDefinition-CommunicateUnstructuredRecommendations.json'));
+const cotestActivity = JSON.parse(readFileSync('fsh-tank/fsh-generated/resources/ActivityDefinition-CotestingScreeningRecommendation.json'));
+const dashboardActivity = JSON.parse(readFileSync('fsh-tank/fsh-generated/resources/ActivityDefinition-DisplayCervicalCancerMedicalHistory.json'));
+const hpvActivity = JSON.parse(readFileSync('fsh-tank/fsh-generated/resources/ActivityDefinition-PrimaryHpvScreeningRecommendation.json'));
+const surveillanceActivity = JSON.parse(readFileSync('fsh-tank/fsh-generated/resources/ActivityDefinition-SurveillanceActivityDefinition.json'));
+const treatmentActivity = JSON.parse(readFileSync('fsh-tank/fsh-generated/resources/ActivityDefinition-TreatmentActivityDefinition.json'));
+
+// Questionnaire resources
+const provideMoreInformationQuestionnaire = JSON.parse(readFileSync('fsh-tank/fsh-generated/resources/Questionnaire-ProvideMoreInformation.json'));
```
mickohanlon23, line 23: Could we just have a loop that brings in all resources from `fsh-tank/fsh-generated/resources/` at once? If for some reason they can't, even making some sort of path variable could be more DRY and less cumbersome.

apply/screening-example.js
```diff
@@ -0,0 +1,118 @@
...
+  JSON.parse(readFileSync('cql/ScreeningImmunocompromisedLibrary.json')),
+  JSON.parse(readFileSync('cql/ScreeningAverageRiskLibrary.json')),
+  JSON.parse(readFileSync('cql/DisplayCervicalCancerMedicalHistory.json')),
+  JSON.parse(readFileSync('cql/CCSMCommonFunctions.json')),
+  JSON.parse(readFileSync('cql/CDSConnectCommonsforFHIRv401.json')),
+];
+
+// Reformat ELM JSON value set references to match what is expected by the 
+// code service built into the cql execution engine
+const elmJsonDependencies = elmJsonDependencyArray.reduce((acc, elm) => {
```
mickohanlon23, line 43: Should the value set reference reformatting be its own function somewhere? This seems like a step we would have to repeat for every test?

apply/screening-example.js
```diff
@@ -0,0 +1,118 @@
...
+    [elm.library.identifier.id]: elm
+  }
+}, {});
+
+// Read in cached value set JSON
+let valueSetJson = JSON.parse(readFileSync('test/ScreeningAverageRisk/.vscache/valueset-db.json'));
+
+// Reformat value sets to match format expected by code service built into the 
+// cql execution engine.
+valueSetJson = Object.keys(valueSetJson).reduce((acc,vs) => {
```
mickohanlon23, line 65: Same comment as for ELM JSON value set references. I found myself writing a very similar function for the CCSM Dashboard, and it could be nice to have this function handy somewhere.

apply/screening-example.js
```diff
@@ -0,0 +1,118 @@
...
+
+// Concatenate all the resources created by $apply operation
+let outputResources = [
+  CarePlan,
+  RequestGroup,
+  otherResources
+];
+
+// Write them out to a file
+writeFileSync('apply/screeningExampleOutput.json', JSON.stringify(outputResources));
```
mickohanlon23, line 118: This is definitely outside the scope of this PR, but it could be nice to have some sort of testing and/or validation of the output. As is, it is a bit hard to tell if the apply function is "doing the right thing," without expanding the JSON and following a series of references in the output array.

mickohanlon23 (approved):
Approving since all of my comments are just suggestions and could easily be applied in a follow-up PR. Encender looks to be generating the proper output, and matches up with the results from cql-testing. Very impressive!

Issue #2, opened 2021-10-14
CommunicationRequest Payload path


mickohanlon23:
When [populating the payload of the CommunicationRequest](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/blob/main/fsh-tank/input/fsh/CervicalCancerScreeningAverageRisk.fsh#L62), the `dynamicValue.path` element of ActivityDefinitions, PlanDefinitions and their sub-actions should refer to the CommunicationRequest's `"payload[0].contentString"` path, instead of simply `"payload"` (see [CommunicationRequest.payload.content_x_](https://www.hl7.org/fhir/communicationrequest-definitions.html#CommunicationRequest.payload.content_x_)).

This change will depend on the [Encender library](https://github.com/ccsm-cds-tools/encender) being able to support more complex FHIRPath strings within DynamicValues.

PR #3, opened 2021-10-15
Addressing issue #2

dwwinters:


mickohanlon23 (commented):
Should the dependencies in `package.json` also be updated to point to `encender` version `0.1.1`?

PR #8, opened 2022-04-05
Changes to CQL library files to fix errors with list promotion disabled

matthewtiller:
a lot of lines are marked as changed because a space was removed at the end of lines. I don't know why that happened, sorry if it clutters the diff.


cql/ScreeningSymptomaticLibrary.cql
```diff
@@ -119,20 +117,27 @@ define MostRecentNilmCytology:
   Dash.MostRecentNilmCytology
 
 define HasNilmCytologyWithinLast12Months:
-  Exists(Common.LookBack(Dash.MostRecentNilmCytology, 12 months))
+  Exists(Common.LookBack({Dash.MostRecentNilmCytology}, 12 months))
```
mickohanlon23, line 120: Instead of casting this to a list to force it to work with `Exists`, could we do something like this to model it similarly to how we do these lookbacks for other singletons (for example in the `MostRecentCoincidentNilmCytologyAndNegativeHpvReport` expression below)?

```suggestion
  Dash.MostRecentNilmCytology C
  where Common.DiagnosticReportDate(C) within 12 months of Today()
  return
    C is not null
```

mickohanlon23, line 120: Instead of casting this to a list to force it to work with `Exists`, could we do something like this to model it similarly to how we do these lookbacks for other singletons (for example in the `MostRecentCoincidentNilmCytologyAndNegativeHpvReport` expression below)?

```suggestion
  Dash.MostRecentNilmCytology C
  where Common.DiagnosticReportDate(C) within 12 months of Today()
  return
    C is not null
```

mickohanlon23 (changes requested):
Thanks for making these updates! The `Exists` vs. `is not null` changes on singletons definitely improve the quality of the CQL code. The spacing changes should be fine - [Atom automatically strips all trailing whitespace](https://flight-manual.atom.io/using-atom/sections/editing-and-deleting-text/#:~:text=The%20%22Remove%20Trailing%20Whitespace%22%20option,panel%20and%20uncheck%20that%20option.) at the end of each line by default, so that probably happened when you saved the files in your editor. I made a few minor comments in `ScreeningSymptomaticLibrary.cql`, the rest of the files looked good! The one major change I would suggest would be to add the updated ELM counterparts to each of the updated CQL files.

cql/ScreeningSymptomaticLibrary.cql
matthewtiller: We're just looking for a true/false here -- is the most recent Nilm cytology within the past 12 months? Would this be the same logic:
`Common.DiagnosticReportDate(Dash.MostRecentNilmCytology) >= Today() - 12 months`

matthewtiller: ok, we'll go with it

cql/ScreeningSymptomaticLibrary.cql
```diff
@@ -1,14 +1,14 @@
-/*
```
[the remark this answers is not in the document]
matthewtiller: I'll take you up on that offer to regenerate the ELM files. I'm running into a certificate issue with my java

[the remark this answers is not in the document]
mickohanlon23: I added the ELM files in this commit: https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/8/commits/1bc5b9a413dfe3d6f2eeadb32566f9e4c49c0590.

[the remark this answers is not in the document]
mickohanlon23: I also added one more tiny commit (https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/8/commits/8d03e9d33ff58a81789cb5236be9520c08e455f0) that changes the reference to the `MostRecentNilmCytology` expression to use the one already defined in that file instead of the one from the Dashboard.

mickohanlon23 (approved):
Nice. No longer getting any errors when I use the cqframework's CQL extension on VSCode! Refactor looks good, and all tests are passing. I'll go ahead and merge this in.

PR #21, opened 2022-06-03
Packager

dwwinters:
Added a packaging script to make it easier to incorporate updates to the CQL and FHIR resources into the dashboard.

@mickohanlon23 Could you take a quick look and review?

@matthewtiller Could you please also review and then merge this into main?


dwwinters:
This is OBE.

PR #43, opened 2022-08-08
Bump FHIR Helpers to version 4.1.000 to align with updated eCQM standard

mickohanlon23:
Update FHIR Helpers to align with [AU 2022 eCQM content](https://github.com/cqframework/ecqm-content-r4-2022/blob/master/input/cql/FHIRHelpers.cql).


mickohanlon23:
I will close this PR as it is not relevant now. If this new FHIR Helpers version is eventually needed for processing CCSM eCQMs, we can revisit. I will keep this branch open.

PR #50, opened 2023-04-18
Updating encender dev dependency and apply example

dwwinters:
Didn't update the version because this only impacts the development examples of the $apply operation.


mickohanlon23 (approved):
Bumps encender to version 0.7.0. Example outputs now include documentation information. Running `npm run screening-example` generates the same output as is committed here, only with different `id`s and dates.

Issue #61, opened 2023-05-18
Shorten Recommendation Text


mickohanlon23:
Update recommendations coming back from CDS to increase brevity and accuracy according to recent walk-throughs with pilot partner clinicians.

Update associated tests.

mickohanlon23:
Branch: [recommendation-edits](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/tree/recommendation-edits)

mickohanlon23:
This was added in https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/64 by @mrnosal. Thanks, Mike!

PR #64, opened 2023-05-25
Recommendation edits

mrnosal:
Incorporate additional feedback on shortening recommendation text for #61 
Incorporate May 23 feedback from RP and RG
Changing default behavior in the CQL to return a list of recommendation strings, not just a single string. 
Updated and ran tests for new recommendation text.


mickohanlon23:
@mrnosal 
I added a few commits:
- Removed two CQL expressions that contained text for recommendations that were now unused
- Bump version number
- Update screening example output, which changed slightly because of how you changed the default behavior in the CQL to return a list of recommendation strings.
-  Ran `npm packager` to create the bundled versions of the ELM and FHIR Resources.

Would you like to take a quick look before I merge into the `main` branch?

mickohanlon23 (approved):

Issue #66, opened 2023-05-26
Expand on text for post-hysterectomy language


mickohanlon23:
We have discussed adding a caveat to all recommendations that account for a hysterectomy. This would expand upon the text to include language stating that you should stop screening only if you have not had a diagnosis of CIN2 or greater in the past 25 years, since this information may not be accessible to the CDS.

Something like:

> “This patient has documentation of a hysterectomy. If hysterectomy was performed for treatment, patients should have 3 consecutive annual HPV-based tests before entering long term surveillance. Long term surveillance involves HPV based testing at 3-year intervals for 25 years.
>  
> Screening is generally not recommended after hysterectomy if no previous diagnosis of CIN2 or greater in the preceding 25 years and the cervix was removed at the time of hysterectomy.”

mickohanlon23:
UPDATE: We will also add logic to recognize if a hysterectomy exists in the patient record without a date, and if so, to give the same general post-hysterectomy recommendation similar to the one described above.

mickohanlon23:
This was resolved in Pull Request https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/79

PR #79, opened 2023-06-09
Add support for Hysterectomy without date #66

mrnosal:
Hysterectomy without date #66  - for special population which looks for cervix removal date, allow record of cervix removal without date. Add proposed recommendation text, test case without Hysterectomy procedure date.

### DIFF
--- a/cql/ManageSpecialPopulation.cql
+++ b/cql/ManageSpecialPopulation.cql
@@ -729,7 +729,7 @@ define RecommendationForManagingPatientsAfterHysterectomy:
     ) then 
       {
         short: 'See Details',
-        date: Collate.DateOfMostRecentReport + 3 years,
+        date: Today(),
         group: 'Hysterectomy (K.4)',
         details: {
           'This patient has documentation of a hysterectomy. If hysterectomy was performed for treatment of histologic HSIL (CIN 2 or CIN 3), patients should have 3 consecutive annual HPV-based tests before entering long term surveillance. Long term surveillance involves HPV based testing at 3-year intervals for 25 years.',
--- a/cql/ManageSpecialPopulation.cql
+++ b/cql/ManageSpecialPopulation.cql
@@ -709,7 +709,7 @@ define RecommendationForManagingPatientsAfterHysterectomy:
         date: ToDate(DateK41),
         group: 'Hysterectomy (K.4)',
         details: {
-          'HPV-based testing (cotest or primary hrHPV test) is recommended annually for 3 years after hysterectomy performed for treatment of CIN 2 or CIN 3.'
+          'HPV-based testing (cotest or primary hrHPV test) is recommended annually for 3 years after hysterectomy performed for treatment of histologic HSIL (CIN 2 or CIN 3).'
         }
       }
     when ( // K4.2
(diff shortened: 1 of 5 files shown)


mrnosal:
Updated recommendation text based on feedback from GM.

cql/ManageSpecialPopulation.cql
```diff
@@ -721,7 +722,20 @@ define RecommendationForManagingPatientsAfterHysterectomy:
         date: Collate.DateOfMostRecentReport + 3 years,
         group: 'Hysterectomy (K.4)',
         details: {
-          'HPV-based testing (cotest or primary hrHPV test) is recommended every 3 years for 25 years.'
+          'HPV-based testing (cotest or primary hrHPV test) is recommended every 3 years for 25 years after hysterectomy is performed for treatment of histologic HSIL (CIN 2 or CIN 3).'
+        }
+      }
+    when ( // no hysterectomy date
+      SAR.AbsenceOrRemovalOfCervix
```
mickohanlon23, line 729:
```suggestion
      Rare.AbsenceOrRemovalOfCervix
```
This same function is defined in `ManageRareAbnormality`, which is already "include"ed in this CQL file. Bringing it in from `ManageRareAbnormality` means we don't have to add another dependency to this file, which should ultimately keep this file a little more lightweight when compiled.

cql/ManageSpecialPopulation.cql
```diff
@@ -22,6 +22,7 @@ include DashboardLibrary version '1.0.0' called Dash
 include CollateManagementData version '1.1.0' called Collate
 include ManageRareAbnormality version '1.1.0' called Rare
 include CCSMCommonFunctions version '1.0.0' called CCF
+include ScreeningAverageRiskLibrary version '1.0.0' called SAR
```
mickohanlon23, line 25:
```suggestion
```
This "include" can be deleted if the suggestion below is approved.

mickohanlon23 (changes requested):
Looks good! The recommendation text and the logic corresponds to what was outlined in the recent email titled "_Expand on text for post-hysterectomy language #66_", and the tests show that this line is successfully catching patients who were given a hysterectomy without a date. Approving pending the one minor suggestion to pull in the `AbsenceOrRemovalOfCervix` from the library that is already included in this library.

Added two additional suggestions regarding the other fields that are returned along with the recommendation.

cql/ManageSpecialPopulation.cql
```diff
@@ -721,7 +722,20 @@ define RecommendationForManagingPatientsAfterHysterectomy:
...
         details: {
-          'HPV-based testing (cotest or primary hrHPV test) is recommended every 3 years for 25 years.'
+          'HPV-based testing (cotest or primary hrHPV test) is recommended every 3 years for 25 years after hysterectomy is performed for treatment of histologic HSIL (CIN 2 or CIN 3).'
+        }
+      }
+    when ( // no hysterectomy date
+      SAR.AbsenceOrRemovalOfCervix
+    ) then 
+      {
+        short: 'Primary HPV or Cotest',
```
mickohanlon23, line 732:
```suggestion
        short: 'See Details',
```

I think this makes more sense now, since we are recommending several different things based upon the interpreted clinical scenario. "See Details" is what we write in other sections where multiple recommendations are given.

cql/ManageSpecialPopulation.cql
```diff
@@ -721,7 +722,20 @@ define RecommendationForManagingPatientsAfterHysterectomy:
...
+        date: Collate.DateOfMostRecentReport + 3 years,
```
mickohanlon23, line 733:
```suggestion
        date: Today(),
```
How we want to handle this is up for discussion, but when the date is ambiguous, we tend to default to 'Today()` as the date for the recommendation.

Another example of a similar type of recommendation where a 'plan' is being recommended as opposed to a specific action is: https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/blob/childbearing-age-25-to-50/cql/ManageRareAbnormality.cql#L1068-L1072.

mickohanlon23 (approved):

PR #87, opened 2023-06-23
Add Endocervical Curettage (ECC) procedure to procedure value set

mickohanlon23:
Add Endocervical Curettage (ECC) procedure to PertinentProcedureShortList value set to be used by smart form mapping in ccsm-cds-service


mickohanlon23:
Merging so that this can be used with https://github.com/ccsm-cds-tools/ccsm-cds-service/pull/17.

PR #96, opened 2023-06-27
Broaden language used in hysterectomy screening recommendation text

mickohanlon23:
Update screening with hysterectomy recommendation text to match more broad language that includes surveillance after CIN 2 or CIN3.

In test patient 52, the doctors discussed wanting to still see language about what to do if a patient had a hysterectomy as a result of a high grade precancer result, even if their history does not explicitly say CIN2 or CIN3.


mrnosal (approved):
Approve recommendation text update

Issue #95, opened 2023-06-27
Update r/t clinical lessons learned from the pilot: Find appropriate mapping for ECC "HSIL"


ssebast2:
Since we now have to be able to reason over all ECC results, there is one result that has not yet been mapped (HSIL). Dr. Guido and/or Updike will need to review every piece of logic in our L2 that uses "Histologic HSIL, unspecified" to determine if that would be an appropriate mapping for ECC "HSIL".

ssebast2:
okay, we'll see what they say

mickohanlon23:
According to Issue https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/issues/97, ECC "HSIL" should now be mapped to  **Histologic HSIL, (CIN3)**.

ssebast2:
ECC HSIL should be mapped to Histologic HSIL, CIN3!!

PR #99, opened 2023-07-11
Add HSIL unspecified to Biopsy Interpretation list

mickohanlon23:
Add HSIL unspecified to Biopsy Interpretation list, so that it can be included in the 'most severe' ranking of biopsy results. This would resolve Issue https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/issues/95.


mrnosal (approved):
Reviewed and ran tests

PR #100, opened 2023-07-13
Fix typo for 5 year interval primary HPV recommendation

mrnosal:
There was a typo in the 5 year primary HPV testing recommendation


mickohanlon23 (approved):
Good catch!

Issue #91, opened 2023-06-26
Placeholder: Additional Recommendation Text Updates


mrnosal:
Placeholder issue for late-breaking recommendation text updates before go-live.

mickohanlon23:
Hysterectomy text has been updated so that screening hysterectomy recommendation now includes the agreed upon text that includes surveillance instructions if hysterectomy was done to manage a CIN2 or CIN3 pre-cancer result.

mickohanlon23:
This Issue will now be to add a non-recommendation for when a patient <25 years old has a NILM cytology with HPV+ cotest. This recommendation should be prioritized after all of our current recommendations, and should say something along the lines of:

> Guidelines do not provide guidance for a patient <25 with a NILM and HPV+ cotest. Please use clinical judgement.

mickohanlon23:
- <25 years NILM cytology with HPV+ cotest recommendation text has been added in https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/101
- Other general recommendation text edits, including those around hysterectomy (mentioned above) have been added in https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/112.

I will close this issue, as no other recommendation text updates are critical for go-live.

PR #101, opened 2023-07-14
Nilm hpv under25 recommendation

mrnosal:
Fix for #91 Provide specialized text for case when CDS cannot make recommendation for patients <25 with NILM and HPV+ cotest result. Note this requires a change to H.2 in ManageRareAbnormality.cql
https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/commit/46424b30d5ca1a4af0a12bf966c075f593be1e92#r121651691
Tests in NoRecommendationProvided were run against ManagementLibrary since they are intended to catch a special case of no recommendation, which means they should not be generating recommendations in Special or Rare.

### DIFF
--- a/cql/ManageCommonAbnormality.cql
+++ b/cql/ManageCommonAbnormality.cql
@@ -512,7 +512,7 @@ define Action:
       ErrataRecommendation is not null
     ) then
       ErrataRecommendation
-    when ( // Handle Special Case < 25
+    when ( // Handle Special Case < 25 NILM & hrHPV positive
       AgeInYears() < 25 and
       MostRecentCytologyCotestWasAfterAge21 and
       Dash.MostRecentNilmCytology is not null and
--- a/cql/ManageCommonAbnormality.cql
+++ b/cql/ManageCommonAbnormality.cql
@@ -361,6 +361,10 @@ define PositivePrimaryHpv:
   ) and 
   Collate.MostRecentCytologyCotest is null
 
+define MostRecentCytologyCotestWasAfterAge21:
+  Collate.MostRecentCytologyCotest.date >= Dash.CervicalCancerScreeningAndMedicalHistory.patientInfo.dateOfBirth + 21 years
+
+
 define Action:
   case
     when (
@@ -504,10 +508,23 @@ define Action:
           TreatmentRecommendationText
         }
       }
-     when (
+    when (
       ErrataRecommendation is not null
     ) then
       ErrataRecommendation
+    when ( // Handle Special Case < 25
+      AgeInYears() < 25 and
+      MostRecentCytologyCotestWasAfterAge21 and
+      Dash.MostRecentNilmCytology is not null and
+      Dash.HasRecentPositiveHpv
+    ) then
+      {
+        short: 'See Details',
+        date: Today(),
+        details: {
+          'There is no specific recommendation for a NILM cytology and HPV + in a patient < 25 years of age. Use clinical judgement.'
+        }
+      }
     else
       null
   end
--- a/test/NoRecommendationProvided/cases/Under25NilmHpvCotestBefore21.yml
+++ b/test/NoRecommendationProvided/cases/Under25NilmHpvCotestBefore21.yml
@@ -1,5 +1,5 @@
 ---
-name: Under 25 NILM HPV Positive
+name: Under 25 NILM HPV Positive Cotest Before Age 21
 
 data:
 - 
@@ -18,19 +18,23 @@ data:
   status: amended
   conclusionCode:
   - SNOMEDCT#373887005 Negative for intraepithelial lesion or malignancy (finding)
-  effectiveDateTime: 2021-01-02
+  effectiveDateTime: 2019-01-02
 - 
   resourceType: DiagnosticReport
   code: LOINC#21440-3 Human papilloma virus 16+18+31+33+35+45+51+52+56 DNA [Presence] in Cervix by Probe
   status: final
   conclusionCode:
   - SNOMEDCT#720005005 Human papillomavirus deoxyribonucleic acid test positive, high risk on cervical specimen (finding)
-  effectiveDateTime: 2021-01-02
+  effectiveDateTime: 2019-01-02
 
+# This is meant to fail CDS recommendation
 results:
-  Recommendation:
-    short: 'See Details'
-    date: '2021-06-02'
-    group: 'General Screening (Table 1)'
-    details:
-    - 'There is no specific recommendation for a NILM cytology and HPV + in a patient < 25 years of age. Use clinical judgement.'
+  HasRecentAbnormalScreening: true
+  HasRecommendation: false
+  RareAbnormalityRecommendation: null
+  SpecialPopulationRecommendation: null
+  CommonAbnormalityRecommendation:
+    short: null
+    date: null
+    details: null
+    group: "General Screening (Table 1)"  
--- a/test/NoRecommendationProvided/cases/Under25NilmHpvNegative.yml
+++ b/test/NoRecommendationProvided/cases/Under25NilmHpvNegative.yml
@@ -1,5 +1,5 @@
 ---
-name: Under 25 NILM HPV18 Positive
+name: Under 25 NILM HPV Negative
 
 data:
 - 
@@ -24,13 +24,17 @@ data:
   code: LOINC#21440-3 Human papilloma virus 16+18+31+33+35+45+51+52+56 DNA [Presence] in Cervix by Probe
   status: final
   conclusionCode:
-  - SNOMEDCT#708299006 Deoxyribonucleic acid of Human papillomavirus 18 (substance)
+  - SNOMEDCT#787724008 Human papillomavirus deoxyribonucleic acid test negative (finding)
   effectiveDateTime: 2021-01-02
 
+# This is meant to fail CDS recommendation
 results:
-  Recommendation:
-    short: 'See Details'
-    date: '2021-06-02'
-    group: 'General Screening (Table 1)'
-    details:
-    - 'There is no specific recommendation for a NILM cytology and HPV + in a patient < 25 years of age. Use clinical judgement.'
+  HasRecentAbnormalScreening: false
+  HasRecommendation: false
+  RareAbnormalityRecommendation: null
+  SpecialPopulationRecommendation: null
+  CommonAbnormalityRecommendation:
+    short: null
+    date: null
+    details: null
+    group: "General Screening (Table 1)"  
--- a/test/NoRecommendationProvided/cases/Under25NilmHpvPos.yml
+++ b/test/NoRecommendationProvided/cases/Under25NilmHpvPos.yml
@@ -1,5 +1,5 @@
 ---
-name: Under 25 NILM HPV16 Positive
+name: Under 25 NILM HPV xPositive
 
 data:
 - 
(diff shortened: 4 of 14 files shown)


cql/ManageCommonAbnormality.cql
```diff
@@ -504,10 +508,23 @@ define Action:
...
-     when (
+    when (
       ErrataRecommendation is not null
     ) then
       ErrataRecommendation
+    when ( // Handle Special Case < 25 NILM & hrHPV positive
+      AgeInYears() < 25 and
+      MostRecentCytologyCotestWasAfterAge21 and
+      Dash.MostRecentNilmCytology is not null and
+      Dash.HasRecentPositiveHpv
```
mickohanlon23, line 519: Does the positive HPV have to be a part of the same cotest as the NILM cytology test?

test/NoRecommendationProvided/cases/Under25NilmHpvCotestBefore21.yml
```diff
@@ -0,0 +1,40 @@
+---
+name: Under 25 NILM HPV Positive Cotest Before Age 21
```
mickohanlon23, line 2: Does the cotest have to be after the patient was 21? Or just under 25? For a 20 year old patient with a NILM/HPV+ cotest, why would we want to return a null recommendation as opposed to this more specific "use clinical judgement" recommendation?

test/NoRecommendationProvided/cases/Under25NilmHpvNegative.yml
```diff
@@ -0,0 +1,40 @@
...
+  effectiveDateTime: 2021-01-02
+- 
+  resourceType: DiagnosticReport
+  code: LOINC#21440-3 Human papilloma virus 16+18+31+33+35+45+51+52+56 DNA [Presence] in Cervix by Probe
+  status: final
+  conclusionCode:
+  - SNOMEDCT#787724008 Human papillomavirus deoxyribonucleic acid test negative (finding)
+  effectiveDateTime: 2021-01-02
+
+# This is meant to fail CDS recommendation
```
mickohanlon23, line 30: In the cqlt.yaml file, this test case only points to the ManagementLibrary. However if you ran this against the screening library, I believe that there would be a recommendation from the average risk screening library for this case, since the patient would have had no abnormal results to bring them over to management. Therefore I propose we may want to delete this test case, since this does not actually reflect what we would expect the overall CDS to return, since they _would_ be getting a recommendation from the AverageRiskScreening libary. I think this would be the same for Under25NilmHpvUnkown case, but it would be worth checking by running them against the screening library.

test/NoRecommendationProvided/cases/Under25NilmHpvUknown.yml
```diff
@@ -0,0 +1,40 @@
...
+  effectiveDateTime: 2021-01-02
+- 
+  resourceType: DiagnosticReport
+  code: LOINC#21440-3 Human papilloma virus 16+18+31+33+35+45+51+52+56 DNA [Presence] in Cervix by Probe
+  status: final
+  conclusionCode:
+  - NULL FLAVOR#UNK Unknown
+  effectiveDateTime: 2021-01-02
+
+# This is meant to fail CDS recommendation
```
mickohanlon23, line 30: Same comment as for Under25NilmHpvNegative case.

test/NoRecommendationProvided/cases/Under25NilmHpvPos.yml
```diff
@@ -0,0 +1,37 @@
+---
+name: Under 25 NILM HPV xPositive
```
mickohanlon23, line 2:
```suggestion
name: Under 25 NILM HPV Positive
```

test/NoRecommendationProvided/cases/Under25NilmHpvUknown.yml
```diff
@@ -0,0 +1,40 @@
+---
```
mickohanlon23, line 1: There is a typo in the name of this test.

mickohanlon23 (changes requested):
Some of the comments are phrased more as points for discussion, if you would like to talk about them once you are back.

cql/ManageCommonAbnormality.cql
mrnosal: Yes, this applies to the most recent cotest with NILM/HPV+

test/NoRecommendationProvided/cases/Under25NilmHpvCotestBefore21.yml
mrnosal: The cotest result must be from age >= 21 to <25 to get this recommendation.

test/NoRecommendationProvided/cases/Under25NilmHpvNegative.yml
mrnosal: I wanted a test which would NOT return the NILM/HPV recommendation, demonstrating the effect of a negative HPV. It could be moved as you suggest. Same for below.

cql/ManageCommonAbnormality.cql
mickohanlon23: In that case, I don't see anything associating the two tests on lines 518 and 519 together here. Should we consider using something like (MostRecentNegativeCotest)[https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/blob/main/cql/ScreeningAverageRiskLibrary.cql#L160]?

test/NoRecommendationProvided/cases/Under25NilmHpvCotestBefore21.yml
mickohanlon23: Why is this the case? And where is this documented? Everything I see on https://jira.mitre.org/browse/CCSMCDS-6 and on https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/issues/91 indicates that the patient only has to be under 25. Why does this recommendation require a patient to be older than 21, when most of the other Under 25 management recommendations (https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/blob/main/cql/ManageSpecialPopulation.cql#L199) only require that the patient is < 25?

test/NoRecommendationProvided/cases/Under25NilmHpvNegative.yml
mickohanlon23: I see what you mean. On second thought I think having this test alongside the Under25HpvPos makes sense to show what the difference between the two would be. I am fine with keeping it.

test/NoRecommendationProvided/cases/Under25NilmHpvUknown.yml
mickohanlon23: Similarly to https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/101#discussion_r1279926731, I am fine with keeping this here to show how the NILM HPV positive management output would differ from the NILM HPV negative management output, as well as the NILM HPV Unknown management output.

test/NoRecommendationProvided/cases/Under25NilmHpvUknown.yml
mickohanlon23: Fixed in https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/101/commits/5bb22c01e25ff224a260fc440973038095845da7.

test/NoRecommendationProvided/cases/Under25NilmHpvPos.yml
mickohanlon23: Fixed in https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/101/commits/5bb22c01e25ff224a260fc440973038095845da7.

test/NoRecommendationProvided/cases/Under25NilmHpvUknown.yml
mickohanlon23: Were you planning to update the name of this test file?

mickohanlon23: Were you planning to update the name of this test file?

mickohanlon23: Were you planning to update the name of this test file?

mickohanlon23:
Just re-reviewed.  A few outstanding comments.

test/NoRecommendationProvided/cases/Under25NilmHpvUknown.yml
mrnosal: Updated

cql/ManageCommonAbnormality.cql
mrnosal: We could create a new function to look for both as part of a cotest if required.

test/NoRecommendationProvided/cases/Under25NilmHpvCotestBefore21.yml
mrnosal: I can't find the reference but Ginny communicated that it needed to be >= 21 since there isn't any guideline for tests before age 21. Can follow up with Ginny.

cql/ManageCommonAbnormality.cql
mickohanlon23: I thought you were saying above that the two tests being a part of a cotest _is_ required? I think this would be worth doing in this PR, but open to thoughts if you think otherwise.

test/NoRecommendationProvided/cases/Under25NilmHpvCotestBefore21.yml
mickohanlon23: I just saw your email. I still don't quite understand the reason for the 21 y.o. age minimum, considering the "recommendation" being given here isn't from a guideline, but is rather saying that there _is not_ a guideline recommendation for this scenario. The whole point of this update is to specify the text we give when a patient falls _outside_ of what is covered by the guideline, and as far as I am aware, the recommendation "_There is no specific recommendation for a NILM cytology and HPV + in a patient < 25 years of age. Use clinical judgement_," holds true whether or not the patient is 21+.

Regardless, the difference between this recommendation and the default one that would alternatively be given to <21 patients in this scenario is minor, since they would be told to "use clinical judgement" in either scenario, so I am fine with resolving this if we want to keep that age cutoff as-is.

mickohanlon23: We discussed during a meeting, and have decided to proceed on with giving this recommendation text only when the cotest occurred after a patient turned 21. Patients under 21 will not receive a recommendation from this section. Thus the current testing aligns with what is expected, and this thread can be resolved.

cql/ManageCommonAbnormality.cql
mickohanlon23: I pushed https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/101/commits/ccf1d773759c997295e2161ea85df0e07e6145b8 which should address this. @mrnosal can you review? Once you give the okay, for this change, I'll approve the PR.

mrnosal (commented):
Built and ran tests. I think this looks good.

cql/ManageCommonAbnormality.cql
mickohanlon23: Mike gave the okay over Slack. Closing this thread as this has been resolved.

mickohanlon23 (approved):
All comments have been addressed.

Issue #28, opened 2022-07-13
Update Risk Tables


mickohanlon23:
There have been email conversations with the NCI team that have resulted in updates to the [Management Risk Tables](https://lhncbc.nlm.nih.gov/LHC-research/LHC-projects/image-processing/cervixca.html). Upload the updated risk tables into our repository once they are published, and ensure that:
1. [generate-cql-risk-tables.py](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/blob/main/risk-tables/generate-cql-risk-tables.py) can accurately convert the updated tables to CQL
2. [AutogeneratedTableLookup.cql](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/blob/main/cql/AutogeneratedTableLookup.cql) can accurately read and interpret the updated fields
3. Tests are created or updated, if needed, to ensure that the right recommendations are given

mickohanlon23:
The updated risk tables have been incorporated in pull request https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/102.

PR #102, opened 2023-07-21
Incorporate Updated Risk Tables into CQL

mickohanlon23:
In January 2022 and February 2022, the NCI Risk Table authors sent us two updated versions of the risk tables posted under the Main Tables section of https://lhncbc.nlm.nih.gov/LHC-research/LHC-projects/image-processing/cervixca.html. In this PR, we bring the new risk tables into the `risk-tables/` directory and run the `generate-cql-risk-tables.py` Python script to update `cql/AutogeneratedRiskTables.cql`, so that our CDS can use these new risk tables.

Several modifications are made to `cql/AutogeneratedTableLookup.cql` to support the modified and additional rows that are included in these updates. Tests are also modified and added to support these updated rows.

This PR resolves Issue https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/issues/28 and also solves a critical patient test script failure by changing a management recommendation from `3-year follow-up` to `1-year follow-up` in Table 4.


mrnosal (approved):
built and ran test-cql

Issue #51, opened 2023-05-10
Update InitialIntensiveSurveillancePeriod to only look for 3 tests rather than 4


mickohanlon23:
**Management Surveillance (J.3.1)** 
The recommendation and logic will need to be updated to include the Errata regarding HPV-based testing at 18 and 30 months.

mickohanlon23:
@ssebast2, @gmmeadows can you update the L2 logic after we confirm what the errata is, if we do plan to implement the errata to this section?

ssebast2:
@mickohanlon23 - I don't see any mention of testing at 18 and 30 months in the errata. Once we receive word that the errata is stable, we will delve into this.

ssebast2:
Oops. Now I see it. It is r/t text in a Figure that is being updated. We used the narrative text to build out the logic b/c we found many of the figures inaccurate or glossing over required details. We'll have to review this more carefully when the article is published - which keeps getting pushed back more and more.

mjo4rk:
I will keep this open so that we can track the update more closely once the errata article is published.

ssebast2:
@mickohanlon23 - Good news! Our logic already looks for a total of 3 negative test results. That is because we used the narrative text of the guideline to develop the L2 vs. the figures in the guideline document. 

Note: Figure 7 was incorrect in the 2019 guidelines. It has been corrected in the 2023 errata document.

For your awareness, our CDS recommends HPV-based testing 6 months after treatment for HSIL (see Section J.2.1). After that, you jump to the Risk tables to determine next steps (i.e., Table 5A). If the 6 month result is negative, they get HPV-based testing in 1 year and it is managed by Table 5B (which recommends a 1 year f/u with HPV-based testing). That totals 3 negative tests that would have been performed at 6 mo, 18 mos, and 30 mos.

mickohanlon23:
Email from Sharon:

This was a translation error on the parts of Rose and I (as opposed to an error in the narrative text of the 2019 guidelines). Side note – there was an error in Figure 7 of the 2019 guidelines that pictured requiring a total of 4 tests. Perhaps that is what swayed us to interpret the narrative guidelines as we did.
- We initially interpreted the guidelines as requiring a neg HPV-based test at 6 months, followed by 3 neg, annual HPV-based tests. 
- I now realize that the guidelines only required a neg. HPV-based test at 6 months, followed by 2 neg annual HPV-based tests (at 18 mos and 30 mos).
[...]

PR #105, opened 2023-07-31
Fix Management Surveillance (J.3.1) L2 Error

mickohanlon23:
* Update InitialIntensiveSurveillancePeriod to only look for 3 tests rather than 4
* Add a test case to show that a patient will now be considered done with their InitialIntensiveSurveillancePeriod whether they completed 4 surveillance tests or 3 surveillance tests.

This resolves Issue https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/issues/51.


mrnosal (approved):
Reviewed code. Built and ran tests.

Issue #93, opened 2023-06-27
Update r/t clinical lessons learned from the pilot: Provide care recommendation for Histologic cancer result


ssebast2:
Regardless of what procedure generated the biopsy and result, if histologic cancer is found, the patient should be referred to GYN ONC.

IF histology test FOUND
AND the result = "Histologic cancer"
THEN DISPLAY Refer to Gynecologic Oncology

mickohanlon23:
Logic for this is addressed by PR: https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/108. Mapping of histologic cancer excision results is done in https://github.com/ccsm-cds-tools/ccsm-cds-service/pull/20.

PR #108, opened 2023-08-02
Histologic Cancer recommendation #93

mrnosal:
Added logic to handle ANY instance of Histologic Cancer via biopsy as a special recommendation.
Added to top of ManageSpecialPopulation as this recommendation supersedes ALL others once histologic cancer biopsy is present.
This does NOT include Diagnoses of cancer, only Histologic Cancer via biopsy.
Added all cervical cancers from valueset to special population resources for test. 
OldCancerHistology now triggers this special recommendation.

### DIFF
--- a/cql/ManageSpecialPopulation.cql
+++ b/cql/ManageSpecialPopulation.cql
@@ -61,12 +61,12 @@ context Patient
 define RecommendationForPatientsWithHistologicCancer:
   case
     when (
-      Rare.HistologyInterpretedAsCancer
+      Rare.AnyHistologyInterpretedAsCancer
     ) then
       {
         short: 'Histologic Cancer',
         date: Today(),
-        group: 'Special Population',
+        group: 'Histologic Cancer',
         details: {
           'Refer to Gynecologic Oncology.'
         }
@@ -773,6 +773,7 @@ define Recommendation:
 
 define WhichPopulationMadeTheRecommendation:
   case
+    when RecommendationForPatientsWithHistologicCancer is not null then 6
     when RecommendationForPatientsYoungerThan25 is not null then 5
     when RecommendationForPregnantPatients is not null then 4
     when RecommendationForImmunosuppressedPatients is not null then 3
--- a/cql/ManageSpecialPopulation.cql
+++ b/cql/ManageSpecialPopulation.cql
@@ -56,6 +56,26 @@ context Patient
 //------------------------------------------------------------------------------
 
 /*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+// Special Populations Management of Patients With Histologic Cancer
+
+define RecommendationForPatientsWithHistologicCancer:
+  case
+    when (
+      Rare.HistologyInterpretedAsCancer
+    ) then
+      {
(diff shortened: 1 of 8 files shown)


cql/ManageSpecialPopulation.cql
```diff
@@ -55,6 +55,26 @@ context Patient
 // RECOMMENDATIONS FOR SPECIAL POPULATIONS
 //------------------------------------------------------------------------------
 
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+// Special Populations Management of Patients With Histologic Cancer
+
+define RecommendationForPatientsWithHistologicCancer:
+  case
+    when (
+      Rare.HistologyInterpretedAsCancer
```
mickohanlon23, line 64: Rare.HistologyInterpretedAsCancer only looks for whether the most recent biopsy was histologic cancer, not if there is ANY instance of Histologic Cancer as your comment mentions

cql/ManageSpecialPopulation.cql
```diff
@@ -55,6 +55,26 @@ context Patient
...
+    ) then
+      {
+        short: 'Histologic Cancer',
+        date: Today(),
+        group: 'Special Population',
```
mickohanlon23, line 69: The group here should be something like "Histologic Cancer". If you look at the other "group" identifiers in this file, they all denote which specific section they come from in the guidelines. This is what ends up displaying on the BPA and helps us to track down where a recommendation came from.

cql/ManageSpecialPopulation.cql
```diff
@@ -744,6 +764,7 @@ define RecommendationForManagingPatientsAfterHysterectomy:
 
 define Recommendation:
   Coalesce(
+    RecommendationForPatientsWithHistologicCancer,
```
mickohanlon23, line 767: I am not sure whether it is still used anywhere, since we now have the "group" fields as a part of each recommendation, but we should probably add RecommendationForPatientsWithHistologicCancer to the "WhichPopulationMadeTheRecommendation" expression below (we can call it 6).

cql/ManageSpecialPopulation.cql
mickohanlon23: I think a test patient where histologic cancer was not the most recent biopsy would be helpful to show this.

mickohanlon23 (changes requested):

cql/ManageSpecialPopulation.cql
mrnosal: Changed to look for cancer in any prior biopsy. Added test case.

cql/ManageSpecialPopulation.cql
mrnosal: Changed group to "Histologic Cancer"

cql/ManageSpecialPopulation.cql
mrnosal: Added "6" as the population for Histologic cancer

cql/ManageSpecialPopulation.cql
mickohanlon23: Great! I like the new test case - shows we should be getting the gyn oncology referral recommendation if histologic cancer is present at all in the patient record, regardless of other biopsy results that may be present before or after.

PR #109, opened 2023-08-10
Update AIS Recommendation

mrnosal:
Provide updated AIS recommendations for CCSMCDS-29
Change order of evaluation to prioritize over other histologic results.
Move some CervixRemoval functions from SpecialPopulation to RareAbnormality
Change alias in RareAbnormality from Common to CCF for consistency with SpecialPopulation
Test cases for 25-50, >50 (No Treatment|Treatment)
Remove AIS from RareAbnormalityHighGradeHistology as it is now handled separately.

### DIFF
--- a/cql/ManageRareAbnormality.cql
+++ b/cql/ManageRareAbnormality.cql
@@ -855,9 +855,6 @@ define CytologyInterpretedAsHsil:
 define CervixRemovalDateIsAfterMostRecentAisBiopsy:
   (CervixRemovalDate is not null) and 
   CervixRemovalDate after Collate.MostRecentAisBiopsy.date
-    
-define MostRecentAisBiopsy:
-  Collate.MostRecentAisBiopsy
 
 define MostRecentTreatmentAfterAis:
   MostRecentTreatment P
--- a/cql/ManageRareAbnormality.cql
+++ b/cql/ManageRareAbnormality.cql
@@ -897,7 +897,7 @@ define RecommendationForAisHistologyResults:
         date: Today(),
         group: 'Managing Histology (I.6)',
         details: {
-          'If negative margins on the excisional specimen, a hysterectomy is preferred.',
+          'If negative margins on the diagnostic excisional specimen, a hysterectomy is preferred.',
           'For patients with positive margins on the excisional specimen, re-excision to achieve negative margins is preferred, even if hysterectomy is planned.',
           'For patients with AIS and persistent positive margins for whom additional excisional procedures are not feasible, either a simple or modified radical hysterectomy is acceptable.'
         }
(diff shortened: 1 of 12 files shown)


cql/ManageRareAbnormality.cql
```diff
@@ -824,6 +852,96 @@ define CytologyInterpretedAsHsil:
       return aC in "HSIL"
   )  
 
+define CervixRemovalDateIsAfterMostRecentAisBiopsy:
+  (CervixRemovalDate is not null) and 
+  CervixRemovalDate after Collate.MostRecentAisBiopsy.date
+    
+define MostRecentAisBiopsy:
```
mickohanlon23, line 859: It looks like this isn't actually used, and the one from `Collate` is used in throughout this file instead. To clean things up, it may help to either delete this expression here, or keep this expression and reference the one in this library throughout this file.

cql/ManageRareAbnormality.cql
```diff
@@ -824,6 +852,96 @@ define CytologyInterpretedAsHsil:
...
+  MostRecentTreatmentAfterAis is not null  
+
+// I.6 Management of AIS over 50: Adoption of Society of Gynecologic Oncology Recommendations
+// Severity of AIS should recommend treatment until hysterectomy is performed.
+define RecommendationForAisHistologyResults:
+  case
+    when (
+      AgeInYears() > 50 and
+      Collate.MostRecentAisBiopsy is not null and
+      not CervixRemovalDateIsAfterMostRecentAisBiopsy and
```
mickohanlon23, line 876: This is an interesting requirement to put in here because UPMC has told us they will not be returning hysterectomies with dates, therefore `CervixRemovalDateIsAfterMostRecentAisBiopsy` will always evaluate to false here, even if the patient did have a hysterectomy. That being said, I think this is technically correct, and I don't actually think this would casue any issues, since if a patient had a hysterectomy without a date, they would be getting a recommendation from the Special Population section, which would be evaluated first.

You can go ahead and resolve this if you think we are good to proceed as is, but I just wanted to bring this up.

mickohanlon23 (approved):
This looks great! My two comments were both not critical, so I'm going to approve the PR. After reviewing the comments, would you mind doing any necesary rebasing and then merging this in?

cql/ManageRareAbnormality.cql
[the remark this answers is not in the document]
mickohanlon23: I removed this unused CQL expression in https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/109/commits/9bcaab5447396c5466f216ffc760899a0c466b83.

Issue #62, opened 2023-05-22
"Rarely Screened" Logic


mickohanlon23:

mickohanlon23:
@gmmeadows Can you point me towards the updated L2 logic and recommendation text for the rarely screened case, once it is finalized?

gmmeadows:
The new logic is on pages 70 – 71 [in the L2 ](https://mitre.sharepoint.com/:w:/r/sites/CervicalCancerCDS/Shared%20Documents/CDS%20Development/L2%20Artifacts/Deliverables/[CDC_CC_CDS_L2Development_v5_2.6.23.docx](https://mitre.sharepoint.com/:w:/r/sites/CervicalCancerCDS/Shared%20Documents/CDS%20Development/L2%20Artifacts/Deliverables/CDC_CC_CDS_L2Development_v5_2.6.23.docx?d=wf7c476a6d04d48dcb865725e459dc24d&csf=1&web=1&e=3EolmN)?d=wf7c476a6d04d48dcb865725e459dc24d&csf=1&web=1&e=3EolmN)in a section under Special Populations called “Guidance for expedited Treatment (Box 1. Essential Changes from Prior Management Guidelines, #3)”. After the discussion with Dr. Guido, I removed the first part of the new logic regarding HPV 16+ and HSIL, since the specific scenarios under which this should be considered are so poorly defined. So now it only includes the logic for Rarely screened, so the issue #62 is fine as is. Let me know if you have any questions on the logic. The recommendation text has now been blessed by both Dr. Guido and Carolann. Carolann is fine with adding this logic versus a toggle switch for UMMC.

mrnosal:
Just figured out that this causes all kinds of headaches with existing tests. We will need to think about how we want to handle these changes.

mrnosal:
First pass at rarely screened logic. Some tests fail, and we decide how to handle those. Are they meant to be rarely screened patients?

mrnosal:
Committed the updated-rarely-screened-logic branch and PR 115.

PR #115, opened 2023-08-18
Update Rarely Screened Logic and tests

mrnosal:
New branch to implement the Rarely Screened logic from the L2 in #62 
The logic looks for the absence of any cytology, cotest, hpv report, cervical precancer treatment and biopsy results in the 5y 6mo preceding a HPV/HSIL cotest. 
Age 27 to 50 gets the treatment with pregnancy concerns recommendation, over 50 does not.


cql/CollateManagementData.cql
```diff
@@ -238,6 +242,12 @@ define MostRecentCytologyReport:
   else
     null
 
+define SecondMostRecentCytologyReport:
+  if Count(SortedCytologyReports) > 0 then
```
mickohanlon23, line 246:
```suggestion
  if Count(SortedCytologyReports) > 1 then
```
We probably only want to query the array for the second most recent cytology report if we know that two exist.

cql/ManageCommonAbnormality.cql
```diff
@@ -365,8 +372,84 @@ define PositivePrimaryHpv:
...
 
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+// Guidance for expedited Treatment (rarely screened) (Box 1. Essential Changes from Prior Management Guidelines, #3)
+// Expedited treatment is preferred for never or rarely screened patients with HPV-positive HSIL cytology regardless of HPV genotype.
+
+define RarelyScreenedLookbackInterval:
+  Interval[Collate.MostRecentCytologyCotest.date - Collate.RarelyScreenedInterval, Collate.MostRecentCytologyCotest.date)  
+  
+
+define IsHsilAndHpvPositive:
```
mickohanlon23, line 383:
```suggestion
define RarelyScreenedLookbackInterval:
  Interval[Collate.MostRecentCytologyCotest.date - Collate.RarelyScreenedInterval, Collate.MostRecentCytologyCotest.date)  

define IsHsilAndHpvPositive:
```

cql/ManageCommonAbnormality.cql
```diff
@@ -365,8 +372,84 @@ define PositivePrimaryHpv:
...
+  Collate.MostRecentCytologyCotestResult = 'HSIL+' and
+  HighRiskHPVPositiveResults
+  
+define HighRiskHPVPositiveResults:
+  AnyTrue(
+    (Collate.MostRecentHpvReport.allConclusions) aC 
+      return aC in Dash."High Risk HPV Positive Results"
+  )
+  
+define PreviousCytologyOrSurveillanceIsOldOrAbsent:
```
mickohanlon23, line 394: Instead of checking if:
```
    No Testing exists OR
    Testing exists but it is outside of the RarelyScreenedLooback interval
```
it may be simpler just to check that:
```
    No testing exists within the RarelyScreened Lookback interval
```
I know we had a conversation around how to structure this logic. Was there a reason we formatted it this way?

cql/ManageCommonAbnormality.cql
```diff
@@ -365,8 +372,84 @@ define PositivePrimaryHpv:
...
+    Collate.SecondMostRecentCytologyCotest is not null and 
+    not (Collate.SecondMostRecentCytologyCotest.date included in RarelyScreenedLookbackInterval)
+  ) or
+  ( 
+    Collate.SecondMostRecentHpvReport is not null and
+    not (Collate.SecondMostRecentHpvReport.date included in RarelyScreenedLookbackInterval)
+  )
+
+define IsRarelyScreenedPatient: 
+  AgeInYears() >= 27 and
```
mickohanlon23, line 414: Why does this start at 27 and not 25?

I know this was discussed at some point, but I'm not sure if this was over email, Jira, GH Issues, or elsewhere. Is it because a patient isn't supposed to start screening until 21, so we want to give them 6 years (5 years + 1 year wiggle room) before considering them rarely screened? If that is the case, why does it matter what the patient was _supposed_ to be doing when we're calculating rarely screened?

mickohanlon23 (changes requested):
Nice! I like how the logic is very succinct, and separated from the `Action` expression. And I appreciate you being so thorough with the amount of test cases!

I left a few comments, but this is very close to being ready to be merged.

cql/ManageCommonAbnormality.cql
mrnosal: This was discussed over email with Ginny. She decided on 27. See email from 8/15 re: Rarely Screened.

cql/ManageCommonAbnormality.cql
mrnosal: What is the suggested change here?

mickohanlon23: Removing the extra empty line.

cql/ManageCommonAbnormality.cql
mickohanlon23: I don't think this is up to us to decide. Doesn't setting an age limit of 27 years old here directly conflict with the main guideline article, which states:

>  For non-pregnant patients 25 years or older, expedited treatment, defined as treatment without preceding colposcopic biopsy demonstrating CIN 2+, is preferred when the immediate risk of CIN 3+ is ≥60%, and is acceptable for those with risks between 25% and 60%. Expedited treatment is preferred for nonpregnant patients 25 years or older with high-grade squamous intraepithelial lesion (HSIL) cytology and concurrent positive testing for HPV genotype 16 (HPV 16) (i.e., HPV 16–positive HSIL cytology) and never or rarely screened patients with HPV-positive HSIL cytology regardless of HPV genotype.

I still think this is far over-complicating the definition of rarely screened. It would be nice for rarely screened to be an objective state: either they were screened in the last 5.5 years, or they were not. I don't understand why a 25 year old would not be considered rarely screened, simply because they were not required to be screened 5 years ago. They still haven't been screened in the last 5 years, bottom line. And more importantly, I do not see where in the guidelines it mentions that a patient cannot be considered rarely screened until they are 27.

cql/ManageCommonAbnormality.cql
mrnosal: Oh, I see now - yes, you can remove the first clause checking for no testing exists (395-399). It is logically redundant

mickohanlon23: Logic was re-worked in the most recent commit: https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/115/commits/97bfa9d1323caa0e7aa2d414dbf3a57efdef635a. Looks good to me!

mickohanlon23 (approved):
Great. My initial comments were addressed, except for one which is pending discussion from the clinical team and can be changed afterwards if necessary, as it just involves modifying an age range. Logic makes sense and aligns with the definition of rarely screened. Tests show that the new recommendation is successfully being given for patients who classify as rarely screened. Thanks @mrnosal!

mickohanlon23:
I added a commit to add the ELM compilation corresponding to your most recent commit's changes to ManageCommonAbnormality.cql.

PR #116, opened 2023-09-01
CCSM-37 Fix the coding mismatch between DiagnosticReport and Observation

yunwwang:
Some DiagnosticReport do not have Biopsy Report code but still references Biopsy Report Observations.

This fix address this use case.


mickohanlon23:
@yunwwang I added some tests in commit https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/116/commits/83bc4621df775fdc21d711b6bf321a9d107359ff to replicate the format of the UMMC biopsy results. Would you mind taking a look? If all looks good, I will go ahead and approve.

mickohanlon23:
I think we need to updated DashboardLibrary.cql to reflect this new way of collating tests. Currently, the dashboard library only returns tests if they are represented as diagnostic reports, and not if they are represented as observations.

Since this will not affect UMMC (they do not display the dashboard output in their CDS app) and should not regress UPMC in any way (all of their labs come across as DiagnosticReports), this should be okay for now, but we will want to make sure this is updated sooner rather than later.

@yunwei - Unless you think this is an easy fix that can be done by tomorrow, would you mind adding a Jira ticket for this at the top of our backlog?

mickohanlon23:
> I think we need to updated DashboardLibrary.cql to reflect this new way of collating tests. Currently, the dashboard library only returns tests if they are represented as diagnostic reports, and not if they are represented as observations.
> 
> Since this will not affect UMMC (they do not display the dashboard output in their CDS app) and should not regress UPMC in any way (all of their labs come across as DiagnosticReports), this should be okay for now, but we will want to make sure this is updated sooner rather than later.
> 
> @yunwwang - Unless you think this is an easy fix that can be done by tomorrow, would you mind adding a Jira ticket for this at the top of our backlog?

I opened [CCSMCDS-61](https://jira.mitre.org/browse/CCSMCDS-61) to address this.

mickohanlon23:
> @yunwwang I added some tests in commit [83bc462](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/commit/83bc4621df775fdc21d711b6bf321a9d107359ff) to replicate the format of the UMMC biopsy results. Would you mind taking a look? If all looks good, I will go ahead and approve.

@yunwwang approved over a recent meeting that the additional tests look good.

mickohanlon23 (approved):

PR #118, opened 2023-09-26
Cin2 surveillance

mrnosal:
Addresses JIRA ticket 38 CRITICAL UPDATE: Incorporate logic to support surveillance of CIN2 (Sections I.2, I.3., K.1 Heading 4) as described in the working L2 v6 draft.

### DIFF
--- a/cql/CollateManagementData.cql
+++ b/cql/CollateManagementData.cql
@@ -387,6 +387,27 @@ define Cin2orCin3Biopsies:
 define MostRecentCin2orCin3Biopsy:
   First(Cin2orCin3Biopsies)
 
+define Cin3Biopsies:
+  SortedBiopsyReports S
+    where S.riskTableInput in {'CIN3'}
+
+define MostRecentCin3Biopsy:
+  First(Cin3Biopsies)
+      
+define Cin2Biopsies:
+  SortedBiopsyReports S
+    where S.riskTableInput in {'CIN2'}
+
+define MostRecentCin2Biopsy:
+  First(Cin2Biopsies)
+  
+define HsilBiopsies:
+  SortedBiopsyReports S
+    where S.riskTableInput in {'HSIL'}
+
+define MostRecentHsilBiopsy:
+  First(HsilBiopsies)
+  
 // Find AIS biopsy
 define AisBiopsies:
   SortedBiopsyReports S
--- a/cql/ManageRareAbnormality.cql
+++ b/cql/ManageRareAbnormality.cql
@@ -66,6 +66,7 @@ code "Endometrial Stromal Cells": 'ESC' from "LOCAL" display 'Endometrial stroma
 //------------------------------------------------------------------------------
 // PARAMETERS
 //------------------------------------------------------------------------------
+parameter CervicalPrecancerTreatmentLookback default 7 years
 
 /***** BEGIN CDS LOGIC ********************************************************/
 
@@ -644,13 +645,6 @@ define AnyHistologyInterpretedAsCin2:
         aC ~ Dash."CIN2"
   )
 
-define MostSevereHistologyInterpretedAsCin2InPast3Years:
-  AnyTrue(
-    (Collate.SortedBiopsyReports.allConclusions) aC
-      return
-        aC ~ Dash."CIN2"
-  )
-
 define MostRecentCin2BiopsyInPast3Years:
   Coalesce(
     (Collate.SortedBiopsyReports) S
@@ -673,7 +667,7 @@ define MostRecentCin2BiopsyInPast2Years:
     )
   )
 
-define MostRecentHsilBiopsyInPast3Years:
+define MostRecentHsilUnpecifiedBiopsyInPast3Years:
     Coalesce(
     (Collate.SortedBiopsyReports) S
     where (S.date >= Now() - 3 years) and
@@ -684,7 +678,7 @@ define MostRecentHsilBiopsyInPast3Years:
     )
   )
 
- define MostRecentHsilBiopsyInPast2Years:
+ define MostRecentHsilUnpecifiedBiopsyInPast2Years:
     Coalesce(
     (Collate.SortedBiopsyReports) S
     where (S.date >= Now() - 2 years) and
@@ -807,23 +801,10 @@ define InitialCervicalHistologyInterpretedAsUnspecifiedHsilOrCin2InLast18Months:
       )
   )
 
-define InitialCervicalHistologyInterpretedAsUnspecifiedHsilOrCin2InLast6Years:
-  First(
-    (Collate.SortedBiopsyReports) B
-      where B.date >= Now() - 6 years and
-      AnyTrue(
-        (B.allConclusions) aC
-          return
-            aC ~ Dash."HSIL, Unspecified" or
-            aC ~ Dash."CIN2"
-      )
-  )
-
 define InitialCervicalHistologyInterpretedAsUnspecifiedHsilOrCin2InLast31Years:
   First(
     (Collate.SortedBiopsyReports) B
       where B.date >= Now() - 31 years and
-      B.date occurs after Patient.birthDate.value and
       AnyTrue(
         (B.allConclusions) aC
           return
@@ -949,7 +930,7 @@ define SubsequentLowGradeCytologyReportsSinceInitialUnspecifiedHsilOrCin2InLast3
     null
 
 
-define RecentNegativeHSILSurveillanceTests:
+define RecentNegativeHsilSurveillanceTests:
   if (
     (SubsequentLowGradeHistologyReportsSinceInitialUnspecifiedHsilOrCin2InLast31Years is not null) and
     (SubsequentLowGradeCytologyReportsSinceInitialUnspecifiedHsilOrCin2InLast31Years is not null)) then
@@ -968,17 +949,19 @@ define RecentNegativeHSILSurveillanceTests:
     else
       null
 
-define FirstTwoNegativeHSILSurveillanceTests:
-  if (Count(RecentNegativeHSILSurveillanceTests) > 1) then
-    RecentNegativeHSILSurveillanceTests[0] union
-    RecentNegativeHSILSurveillanceTests[1]
-  else
-    null
+define HasTwoNegativeHsilSurveillanceTests:
+  Count(RecentNegativeHsilSurveillanceTests) >= 2
     
-define NegativeSurveillanceTestDatesAfterHSILSurveillance:
-  if (Count(FirstTwoNegativeHSILSurveillanceTests) > 1) then
+define SecondNegativeHsilSurveillanceTestDate:
+  if (HasTwoNegativeHsilSurveillanceTests) then
+      RecentNegativeHsilSurveillanceTests[1].date
+  else 
+    null
+
+define NegativeSurveillanceTestDatesAfterHsilSurveillance:
+  if HasTwoNegativeHsilSurveillanceTests then
     SortedNegativeSurveillanceTests T
-    where T.date after Last(FirstTwoNegativeHSILSurveillanceTests).date
+    where T.date after SecondNegativeHsilSurveillanceTestDate
   else
     null
 
@@ -996,7 +979,7 @@ define CytologyInterpretedAsHsil:
   AnyTrue(
     (MostRecentCytologyReport.allConclusions) aC
       return aC in "HSIL"
-  )  
+  )
 
 define CervixRemovalDateIsAfterMostRecentAisBiopsy:
   (CervixRemovalDate is not null) and 
@@ -1007,7 +990,18 @@ define MostRecentTreatmentAfterAis:
   where CCF.ProcedureDate(P) occurs after Collate.MostRecentAisBiopsy.date
 
 define HasTreatmentAfterAis:
-  MostRecentTreatmentAfterAis is not null  
+  MostRecentTreatmentAfterAis is not null
+
+/**
+* Check if the the date of the most recent cervical precancer treatment occurs
+* starting during the lookback period before the given date until the present time
+* @param d - a System date to compare
+* @param Lookback - a time quantity to be evaluated before date d
+* @returns {Boolean: external}
+*/
+define function HadNoCervicalPrecancerTreatmentFromDateToPresent(d System.DateTime, Lookback System.Quantity):
+  Collate.DateOfLastCervicalPrecancerTreatment is null or 
+  Collate.DateOfLastCervicalPrecancerTreatment in Interval[d - Lookback, Now()]
 
 // I.6 Management of AIS over 50: Adoption of Society of Gynecologic Oncology Recommendations
 // Severity of AIS should recommend treatment until hysterectomy is performed.
@@ -1044,7 +1038,7 @@ define RecommendationForAisHistologyResults:
           'For patients with positive margins on the excisional specimen, re-excision to achieve negative margins is preferred, even if hysterectomy is planned.',
           'For patients with AIS and persistent positive margins for whom additional excisional procedures are not feasible, either a simple or modified radical hysterectomy is acceptable.'
         }
-      }      
+      }
     when (
       AgeInYears() >= 25 and AgeInYears() <= 50 and
       Collate.MostRecentAisBiopsy is not null and
@@ -1117,18 +1111,14 @@ define RecommendationForHistologyResults:
           'Treatment is recommended for histologic HSIL.'
         }
       }
-      
+
 // =====================================================      
     // I.2 Management of Histologic HSIL (CIN2 or CIN3)
     when ( // I2.1
       AgeInYears() >= 25 and
       not Dash.Pregnant and
       AnyHistologyInterpretedAsCin3 and
-      (
-        Collate.DateOfLastCervicalPrecancerTreatment is null or 
-        not (Collate.DateOfLastCervicalPrecancerTreatment occurs after Collate.MostRecentCin3Biopsy.date) or
-        Collate.DateOfLastCervicalPrecancerTreatment 7 years or more before Collate.MostRecentCin3Biopsy.date
-      )
+      HadNoCervicalPrecancerTreatmentFromDateToPresent(Collate.MostRecentCin3Biopsy.date, CervicalPrecancerTreatmentLookback)
     ) then
       {
         short: 'Treatment',
@@ -1144,11 +1134,7 @@ define RecommendationForHistologyResults:
       AgeInYears() >= 25 and AgeInYears() <= 50 and
       not Dash.Pregnant and not RecentlyRespondedYesToFuturePregnancyConcernsQuestion and
       HistologyInterpretedAsCin2 and
-      (
-        Collate.DateOfLastCervicalPrecancerTreatment is null or 
-        not (Collate.DateOfLastCervicalPrecancerTreatment occurs after Collate.MostRecentCin2Biopsy.date) or
-        Collate.DateOfLastCervicalPrecancerTreatment 7 years or more before Collate.MostRecentCin2Biopsy.date
-      )
+      HadNoCervicalPrecancerTreatmentFromDateToPresent(Collate.MostRecentCin2Biopsy.date, CervicalPrecancerTreatmentLookback)
     ) then
       {
         short: 'Surveillance or Treatment',
@@ -1165,11 +1151,7 @@ define RecommendationForHistologyResults:
       AgeInYears() > 50 and
       not Dash.Pregnant and not RecentlyRespondedYesToFuturePregnancyConcernsQuestion and
       AnyHistologyInterpretedAsCin2 and
-      (
-        Collate.DateOfLastCervicalPrecancerTreatment is null or 
-        not (Collate.DateOfLastCervicalPrecancerTreatment occurs after Collate.MostRecentCin2Biopsy.date) or
-        Collate.DateOfLastCervicalPrecancerTreatment 7 years or more before Collate.MostRecentCin2Biopsy.date
-      )
+      HadNoCervicalPrecancerTreatmentFromDateToPresent(Collate.MostRecentCin2Biopsy.date, CervicalPrecancerTreatmentLookback)
     ) then
       {
         short: 'Treatment',
@@ -1180,7 +1162,7 @@ define RecommendationForHistologyResults:
           'Excisional treatment is preferred, and treatment with ablation is acceptable.'
         }
       }
-            
+
     // I.3 Management of CIN2 in Those Who Are Concerned About the Potential Effect of Treatment on Future Pregnancy Outcomes
     when ( // I3.5
       AgeInYears() >= 25 and AgeInYears() <= 50 and
@@ -1190,11 +1172,7 @@ define RecommendationForHistologyResults:
       ) and
       HasHistologyInterpretedAsUnspecifiedHsilOrCin2BeforeMostRecentBiopsy and
       RecentlyRespondedYesToFuturePregnancyConcernsQuestion and
-      (
-        Collate.DateOfLastCervicalPrecancerTreatment is null or 
-        not (Collate.DateOfLastCervicalPrecancerTreatment occurs after Collate.MostRecentCin2Biopsy.date) or
-        Collate.DateOfLastCervicalPrecancerTreatment 7 years or more before Collate.MostRecentCin2Biopsy.date
-      )
+      HadNoCervicalPrecancerTreatmentFromDateToPresent(Collate.MostRecentCin2Biopsy.date, CervicalPrecancerTreatmentLookback)
     ) then
       {
         short: 'Surveillance or Treatment',
@@ -1204,12 +1182,12 @@ define RecommendationForHistologyResults:
           'Treatment is recommended for histologic HSIL (CIN 2) or histologic HSIL (unspecified) lasting 2 years.'
         }
       }
-  
+
     when ( // I3.4
       AgeInYears() >= 25 and AgeInYears() <= 50 and
       (InitialCervicalHistologyInterpretedAsUnspecifiedHsilOrCin2InLast31Years is not null) and
-      Count(FirstTwoNegativeHSILSurveillanceTests) >= 2 and
-      Count(NegativeSurveillanceTestDatesAfterHSILSurveillance) >= 3 and
+      HasTwoNegativeHsilSurveillanceTests and
+      Count(NegativeSurveillanceTestDatesAfterHsilSurveillance) >= 3 and
       RecentlyRespondedYesToFuturePregnancyConcernsQuestion
     ) then
       {
@@ -1238,13 +1216,9 @@ define RecommendationForHistologyResults:
       }
     when ( // I3.2
       AgeInYears() >= 25 and AgeInYears() <= 50 and
-      MostRecentHsilBiopsyInPast3Years is not null and
+      MostRecentHsilUnpecifiedBiopsyInPast3Years is not null and
       RecentlyRespondedYesToFuturePregnancyConcernsQuestion and
-      (
-        Collate.DateOfLastCervicalPrecancerTreatment is null or 
-        not (Collate.DateOfLastCervicalPrecancerTreatment occurs after Collate.MostRecentHsilBiopsy.date) or
-        Collate.DateOfLastCervicalPrecancerTreatment 7 years or more before Collate.MostRecentHsilBiopsy.date
-      )
+      HadNoCervicalPrecancerTreatmentFromDateToPresent(Collate.MostRecentHsilBiopsy.date, CervicalPrecancerTreatmentLookback)
     ) then
       {
         short: 'Surveillance or Treatment',
@@ -1259,11 +1233,7 @@ define RecommendationForHistologyResults:
       AgeInYears() >= 25 and AgeInYears() <= 50 and
       MostRecentCin2BiopsyInPast3Years is not null and
       RecentlyRespondedYesToFuturePregnancyConcernsQuestion and
-      (
-        Collate.DateOfLastCervicalPrecancerTreatment is null or 
-        not (Collate.DateOfLastCervicalPrecancerTreatment occurs after Collate.MostRecentCin2Biopsy.date) or
-        Collate.DateOfLastCervicalPrecancerTreatment 7 years or more before Collate.MostRecentCin2Biopsy.date
-      )
+      HadNoCervicalPrecancerTreatmentFromDateToPresent(Collate.MostRecentCin2Biopsy.date, CervicalPrecancerTreatmentLookback)
     ) then
       {
         short: 'Surveillance or Treatment',
@@ -1512,12 +1482,6 @@ define RecommendationForHistologyResults:
       null
   end
 
-define bothTreatmentBiopsy:
-      (
-        Collate.DateOfLastCervicalPrecancerTreatment is null or 
-        not (Collate.DateOfLastCervicalPrecancerTreatment occurs after Collate.MostRecentCin3Biopsy.date)
-      )
-        
 /*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 // RARE ABNORMALITY #4: Surveillance After Abnormalities
 
@@ -1759,6 +1723,11 @@ define InitialIntensiveSurveillancePeriod:
   else
     false
 
+define HasHistologyAfterTreatmentForHighGradeResult:
+  Collate.MostRecentBiopsyReport is not null and
+  TreatmentForHighGradeHistologyOrCytology is not null and
+  Collate.MostRecentBiopsyReport.date after TreatmentForHighGradeHistologyOrCytologyDate
+  
 define RecommendationForSurveillanceAfterAbnormalities:
   case
     // J.2 Short-Term Follow-up After Treatment for Histologic HSIL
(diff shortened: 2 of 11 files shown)


cql/ManageRareAbnormality.cql
```diff
@@ -613,6 +613,9 @@ define RecommendationForExceptionsToColposcopyThreshold:
 define MostRecentBiopsyReportWasWithinPastYear:
   Collate.MostRecentBiopsyReport.date >= Now() - 1 year
 
+define MostRecentBiopsyReportWasWithinPastThreeYears:
```
mickohanlon23, line 616: This appears to be unused throughout the CQL logic.

cql/ManageRareAbnormality.cql
```diff
@@ -627,6 +630,70 @@ define HistologyInterpretedAsCin3:
...
+  )
+
+define AnyHistologyInterpretedAsCin2:
+  AnyTrue(
+    (Collate.SortedBiopsyReports.allConclusions) aC
+      return
+        aC ~ Dash."CIN2"
+  )
+
+define MostSevereHistologyInterpretedAsCin2InPast3Years:
```
mickohanlon23, line 647: 1. This name of this expression includes "InPast3Years", but there is no 3 year lookback period in the definition
2. Is this unused? One place it could be used is within the MostRecentCin2BiopsyInPast3Years definition below. Otherwise, I think it can be deleted.

cql/ManageRareAbnormality.cql
```diff
@@ -627,6 +630,70 @@ define HistologyInterpretedAsCin3:
...
+  AnyTrue(
+    (Collate.SortedBiopsyReports.allConclusions) aC
+      return
+        aC ~ Dash."CIN2"
+  )
+
+define MostRecentCin2BiopsyInPast3Years:
```
mickohanlon23, line 654: I believe this should be named "Any" instead of "MostRecent".

cql/ManageRareAbnormality.cql
```diff
@@ -627,6 +630,70 @@ define HistologyInterpretedAsCin3:
...
+  Coalesce(
+    (Collate.SortedBiopsyReports) S
+    where (S.date >= Now() - 3 years) and
+    AnyTrue(
+    (S.allConclusions) aC
+      return 
+        aC ~ Dash."CIN2"
+    )
```
mickohanlon23, line 662: It looks like the AnyHistologyInterpretedAsCin2 expression you defined above already does this. Consider using that expression in these follow-on expressions, so that you don't have to keep redefining the same logic.

cql/ManageRareAbnormality.cql
```diff
@@ -627,6 +630,70 @@ define HistologyInterpretedAsCin3:
...
+define MostRecentCin2BiopsyInPast2Years:
+  Coalesce(
+    (Collate.SortedBiopsyReports) S
+    where (S.date >= Now() - 2 years) and
+    AnyTrue(
+    (S.allConclusions) aC
+      return 
+        aC ~ Dash."CIN2"
+    )
+  )
```
mickohanlon23, line 674: Same 2 comments from `MostRecentCin2BiopsyInPast3Years` expression above apply to `MostRecentCin2BiopsyInPast2Years`.

cql/ManageRareAbnormality.cql
```diff
@@ -627,6 +630,70 @@ define HistologyInterpretedAsCin3:
...
+ define MostRecentHsilBiopsyInPast2Years:
+    Coalesce(
+    (Collate.SortedBiopsyReports) S
+    where (S.date >= Now() - 2 years) and
+    AnyTrue(
+    (S.allConclusions) aC
+      return 
+        aC ~ Dash."HSIL, Unspecified"
+    )
+  )
```
mickohanlon23, line 696: Both of these expressions (`MostRecentHsilBiopsyInPast3Years` and  `MostRecentHsilBiopsyInPast2Years`) should be named "Any" instead of "MostRecent" to avoid confusion, since these definitions do not look for the most recent biopsy, but rather any biopsy that has a value of "HSIL, Unspecified"

mickohanlon23, line 696: Both of these expressions (`MostRecentHsilBiopsyInPast3Years` and  `MostRecentHsilBiopsyInPast2Years`) should have the name HSIL unspecified in the title rather than just HSIL, since an HSIL biopsy technically refers to a CIN2 or a CIN3, whereas this is an HSIL biopsy that is specifically not specified between the two, and has different recommendations associated with it.

cql/ManageRareAbnormality.cql
```diff
@@ -752,11 +819,24 @@ define InitialCervicalHistologyInterpretedAsUnspecifiedHsilOrCin2InLast6Years:
       )
   )
 
+define InitialCervicalHistologyInterpretedAsUnspecifiedHsilOrCin2InLast31Years:
```
mickohanlon23, line 822: The expression above this - `InitialCervicalHistologyInterpretedAsUnspecifiedHsilOrCin2InLast6Years` - should be removed if it is now unused.

cql/ManageRareAbnormality.cql
```diff
@@ -752,11 +819,24 @@ define InitialCervicalHistologyInterpretedAsUnspecifiedHsilOrCin2InLast6Years:
...
+  First(
+    (Collate.SortedBiopsyReports) B
+      where B.date >= Now() - 31 years and
+      B.date occurs after Patient.birthDate.value and
```
mickohanlon23, line 826: I'm wondering if this adds anything? If so, why would we check that the test happens after the patient was born here, and not anywhere else?

cql/ManageRareAbnormality.cql
```diff
@@ -844,6 +948,40 @@ define SubsequentLowGradeCytologyReportsSinceInitialUnspecifiedHsilOrCin2InLast6
   else
     null
 
+
+define RecentNegativeHSILSurveillanceTests:
```
mickohanlon23, line 952: `HSIL` -> `Hsil` for consistency. Same with `FirstTwoNegativeHSILSurveillanceTests` and `NegativeSurveillanceTestDatesAfterHSILSurveillance` expressions below.

mickohanlon23 (changes requested):
Part 1 of Review: ManageRareAbnormality.cql CQL expression definitions to support section I.2 and I.3, before actual recommendations are introduced.

cql/ManageRareAbnormality.cql
```diff
@@ -980,12 +1117,18 @@ define RecommendationForHistologyResults:
...
       AgeInYears() >= 25 and
       not Dash.Pregnant and
-      MostRecentBiopsyReportWasWithinPastYear and
-      HistologyInterpretedAsCin3
+      AnyHistologyInterpretedAsCin3 and
+      (
+        Collate.DateOfLastCervicalPrecancerTreatment is null or 
+        not (Collate.DateOfLastCervicalPrecancerTreatment occurs after Collate.MostRecentCin3Biopsy.date) or
+        Collate.DateOfLastCervicalPrecancerTreatment 7 years or more before Collate.MostRecentCin3Biopsy.date
+      )
```
mickohanlon23, line 1131:
```suggestion
(Collate.DateOfLastCervicalPrecancerTreatment occurs on or after (Collate.MostRecentCin3Biopsy.date - 7 years)) is null or false
```
Consider this one-liner, or something like it. May make things a bit more concise and readable.

cql/ManageRareAbnormality.cql
```diff
@@ -1000,8 +1143,12 @@ define RecommendationForHistologyResults:
     when ( // I2.2
       AgeInYears() >= 25 and AgeInYears() <= 50 and
       not Dash.Pregnant and not RecentlyRespondedYesToFuturePregnancyConcernsQuestion and
-      MostRecentBiopsyReportWasWithinPastYear and
-      HistologyInterpretedAsCin2
+      HistologyInterpretedAsCin2 and
```
mickohanlon23, line 1146: Shouldn't this be the `MostRecentCin2BiopsyInPast3Years` expression (which doesn't actually look for the most recent biopsy but rather any in the past 3 years that are CIN2)

cql/ManageRareAbnormality.cql
mickohanlon23, line 1131: Whether or not this logic is condensed according to the logic I suggested above, this logic should be abstracted out to it's own CQL expression. We should make some sort of "NotTreated" expression that includes the logic statements to make things more efficient, since it is repeated many times throughout section.

You could even get fancy and make it a function, that takes in a biopsy date as the parameter value, and returns true if there is no treatment 7 years before or anytime after that date. Doing something like this would make the actual recommendation logic more readable and would make it so that you don't have to change several different parts of the code if one piece of this logic needs to be tweaked.

cql/ManageRareAbnormality.cql
```diff
@@ -1038,23 +1189,27 @@ define RecommendationForHistologyResults:
...
+  
     when ( // I3.4
       AgeInYears() >= 25 and AgeInYears() <= 50 and
-      (InitialCervicalHistologyInterpretedAsUnspecifiedHsilOrCin2InLast6Years is not null) and
-      Count(SubsequentLowGradeHistologyReportsSinceInitialUnspecifiedHsilOrCin2InLast6Years) = 5 and
-      Count(SubsequentLowGradeCytologyReportsSinceInitialUnspecifiedHsilOrCin2InLast6Years) >= 5 and
+      (InitialCervicalHistologyInterpretedAsUnspecifiedHsilOrCin2InLast31Years is not null) and
+      Count(FirstTwoNegativeHSILSurveillanceTests) >= 2 and
+      Count(NegativeSurveillanceTestDatesAfterHSILSurveillance) >= 3 and
       RecentlyRespondedYesToFuturePregnancyConcernsQuestion
```
mickohanlon23, line 1213: Sharon added a final line to this recommendation:

> AND the patient has only received ‘negative surveillance tests’ AFTER the above surveillance criteria


 as well as the following comment, which I think needs to be addressed here:

 

> "_Technical team - this is trying to say that as long as no abnormal results occurred after the initial intense surveillance period, the patient can continue to be screened every 3 years for 25 years._"

cql/ManageRareAbnormality.cql
```diff
@@ -1346,6 +1512,12 @@ define RecommendationForHistologyResults:
       null
   end
 
+define bothTreatmentBiopsy:
```
mickohanlon23, line 1515: Should this be removed? It looks like it's not used anywhere.

cql/ManageRareAbnormality.cql
```diff
@@ -1635,11 +1807,15 @@ define RecommendationForSurveillanceAfterAbnormalities:
...
+      
+// =============================================================================      
     // J.3 Guidance for Long-Term Follow-up After Treatment for High-Grade Histology or Cytology
     when ( // J3.1
       HasTreatmentForHighGradeHistologyOrCytology and
       TreatmentForHighGradeHistologyOrCytologyDate >= Now() - 25 years and
-      not InitialIntensiveSurveillancePeriod
+      not InitialIntensiveSurveillancePeriod // AFTER the treatment for high-grade HistologyOrCytology
+      // AND NOT Cervical Histology result AFTER the 'treatment for high-grade HistologyOrCytology
+      // Update aug 28 - sharon assigned task to change this
```
mickohanlon23, line 1818: Did we intend to commit this? If so, can we mark it as "TODO: *description of pending update*", similarly to how we do elsewhere in the CQL code?

mickohanlon23 (changes requested):
Review Pt 2: Section I.2 and I.3 recommendation logic statements

cql/CollateManagementData.cql
```diff
@@ -387,6 +387,27 @@ define Cin2orCin3Biopsies:
 define MostRecentCin2orCin3Biopsy:
   First(Cin2orCin3Biopsies)
 
+define Cin3Biopsies:
```
mickohanlon23, line 390: It seems like these functions repeat a lot of what is defined in ManageRareAbnormality.cql. For example, it seems like `CIN2Biopsies` is essentially the same as `Rare.HistologyInterpretedAsCin2`. I know there are some subtle nuances, like the difference between "the most recent CIN2 histology result" and "the most recent histology result is CIN2", so I'm not entirely sure if this logic was intended to live as it's own logic that is separate from the ManageRareAbnormality.cql file, but it seems like some if not most of what is here can be consolidated with what lives over there.

cql/ManageSpecialPopulation.cql
```diff
@@ -330,8 +330,11 @@ define RecommendationForPatientsYoungerThan25:
       }  
     when ( // Heading4.1
       AgeInYears() < 25 and
-      Rare.MostRecentBiopsyReportWasWithinPastYear and
-      Rare.HistologyInterpretedAsCin3
+      Rare.AnyHistologyInterpretedAsCin3 and
+      (
+        Collate.DateOfLastCervicalPrecancerTreatment is null or 
+        not (Collate.DateOfLastCervicalPrecancerTreatment occurs after Collate.MostRecentCin3Biopsy.date)
+      )
```
mickohanlon23, line 337: Kind of like in section I, it could be nice to create an expression called something like "NoTreatmentAfterCIN3", to be used here as well as in the rest of the expressions in K.1 Heading 4. Or better yet a function called "NoTreatmentAfterDate" that does the same thing but works for any date inputted as a parameter.

cql/ManageRareAbnormality.cql
```diff
@@ -844,6 +948,40 @@ define SubsequentLowGradeCytologyReportsSinceInitialUnspecifiedHsilOrCin2InLast6
...
+         (SubsequentLowGradeHistologyReportsSinceInitialUnspecifiedHsilOrCin2InLast31Years) H 
+           with (NegativeHrHpvTests) N
+             such that N.date same as H.date 
+            )
+      ) R 
+      sort by date asc
+    else
+      null
+
+define FirstTwoNegativeHSILSurveillanceTests:
```
mickohanlon23, line 971: I have a CQL VSCode extension throwing the following error:
```
Could not resolve call to operator Union with signature (tuple{riskTableInput:System.String,allConclusions:list<FHIR.CodeableConcept>,date:System.DateTime},tuple{riskTableInput:System.String,allConclusions:list<FHIR.CodeableConcept>,date:System.DateTime}).
```
I think you are only supposed to call a union on two lists, but the CQL-to-ELM translator just figures out how to union two singletons if they have the same signature.

mickohanlon23: It seems like all you actually need from this expression is:
* A boolean determining whether or not at least 2 negative HSIL surveillance tests exist
* The date of the second test

I propose instead of having this expression returning the two tests themselves, you could have it return the date of the second test if the count is >= 2, or else return null. That way you could say "if FirstTwoNegativeHSILSurveillanceTests is not null" every time you are trying to evaluate whether the two tests have happened yet, and then when you need the date, you just call this expression.

mickohanlon23 (changes requested):
Review pt 3: Section K.1 Changes in ManageSpecialPopulation.cql, Expressions added to CollateManagementData.cql, and warning being thrown in ManageRareAbnormality.cql.

These are the last of the comments I had for my initial review. I split them up into sections so that my comments didn't accidentally all delete from me on GitHub. Let me know if you would like to discuss anything.

As a whole this looks awesome and seems to very tightly reflect the L2, nice job!

cql/ManageRareAbnormality.cql
mrnosal: This function is used in the updated I.2.2 which is part of[ PR 120](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/120).

cql/ManageRareAbnormality.cql
mrnosal: Agree. This can be deleted.

cql/ManageRareAbnormality.cql
mrnosal: It is using the SortedBiopsyReports list, ordered newest to oldest  and Coalesce, so the first match returned will be the most recent biopsy report. Same logic used for other "MostRecentXXXInPastZYears" definitions.

cql/ManageRareAbnormality.cql
mrnosal: This is iterating over a sorted array, and stopping at the first match, so it is returning the most recent biopsy. AnyHistologyInterpretedAsCin2 will look through all results for CIN2.

cql/ManageRareAbnormality.cql
mrnosal: This is a common pattern to iterate over a sorted list to find the 'most recent' result of type X, so the definitions are as intended.

cql/ManageRareAbnormality.cql
mrnosal: These are iterating over the SortedBiopsyReports list, which goes from newest to oldest. Coalesce returns the first match, which is the 'most recent' of the matches.

cql/ManageRareAbnormality.cql
mrnosal: Yes, this was changed to 31 years so this expression can be removed.

cql/ManageRareAbnormality.cql
mrnosal: I think this was something I was using during debugging, so this can be removed now.

cql/ManageRareAbnormality.cql
mrnosal: Agree - I'll make the names consistent.

cql/ManageRareAbnormality.cql
mrnosal: Good suggestion. Will update the logic in other conditionals to use this.

cql/ManageRareAbnormality.cql
mrnosal: MostRecentCin2BiopsyInPast3Years does return only the MRB, no any. This section is updated to AnyHistologyInterpretedAsCin2 in the [PR 120](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/120) anyway.

cql/ManageRareAbnormality.cql
mrnosal: Agree - will update the expression name to HsilUnspecified

cql/ManageRareAbnormality.cql
mrnosal: Not used - will remove.

cql/ManageRareAbnormality.cql
mrnosal: Will add update to check for histology after treatment

cql/CollateManagementData.cql
mrnosal: They are different functions and have different syntactical usage in the CQL. We could consider where such definitions belong and whether we need to consolidate. We also need to consider the need to prefix all these references if moved (e.g. HistologyInterpretedAsCin2 becomes Collate.HistologyInterpretedAsCin2 in usage)

cql/ManageSpecialPopulation.cql
mrnosal: Will update.

cql/ManageRareAbnormality.cql
mrnosal: Will refactor to a boolean for HasTwoNegativeHsilSurveillanceTests and date of second test SecondNegativeHsilSurveillanceTestDate

mrnosal:
All comments should be addressed with c270489

cql/ManageRareAbnormality.cql
mickohanlon23: You are filtering the SortedBiopsyReports list by only those reports that have a conclusionCode of CIN2. So the first match returned will not be the most recent biopsy report. It will be the most recent report that has a value of CIN2. This expression is doing the same thing as `AnyHistologyInterpretedAsCin2`, except adding a date range, and taking the most recent of the CIN2 results instead of returning `true`.

If the patient had a CIN2 and then CIN1 afterwards, this expression would still return the CIN2, even if it was not the most recent biopsy.

Compare this to "[HistologyInterpretedAsCin2](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/blob/cin2-surveillance/cql/ManageRareAbnormality.cql#L620)", which looks for the most recent biopsy, and then evaluates whether or not that biopsy is CIN2. _Technically_ the name of this expression is not incorrect, because we are looking for the most recent CIN2 biopsy here. But I think now that we have expressions that look for _any_ biopsy over a given span, as well as expressions that look for the _most recent_ biopsy over a given span, we need to be clear and consistent with how we label most recent. And if I see an expression titled "Most Recent..." I am thinking that it is looking at the most recent biopsy that a patient had, regardless of result.

cql/ManageRareAbnormality.cql
mickohanlon23: I think you need to take the order of operations here. `Coalesce` is called on the outside of this expression and is the last thing to evaluate. This is not returning the first entry of `SortedBiopsyReports`. This is returning the first entry of a subset of that array, after you have:
1. Filtered by date
2. Filtered by only biopsy reports that have a result of CIN2. 

Coalesce is called after that sorting has been done, so you are returning the most recent of the CIN2 biopsy results in the past 3 years, even if it was not the most recent overall biopsy report.

cql/ManageRareAbnormality.cql
mickohanlon23: This has been removed in the most recent commit.

cql/ManageRareAbnormality.cql
mickohanlon23: This was removed in the most recent commit. All tests still pass.

cql/ManageRareAbnormality.cql
mickohanlon23: I still would like to discuss this, however, as you said, I see that I opened another comment regarding this I2.2 section in PR 120: https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/120#discussion_r1379480063 - so we can continue the discussion there. I will resolve this, since this will be addressed in https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/120.

cql/ManageRareAbnormality.cql
mickohanlon23: This is unused throughout the CQL, and was removed in the most recent commit.

cql/ManageRareAbnormality.cql
mickohanlon23: They are iterating over a list of "HSIL, Unspecified" results, not a list of all biopsy results. This is because of the `where` clause you have here. `Coalesce` is being called after you do the filtering with the `where` clause. I can't quite tell by your comment whether that was intentional or not, but this does not seem to align with the use of "Most Recent" in other parts of our CQL logic, where we take take the most recent test from the unfiltered list of SortedBiopsyReports. See https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/118#discussion_r1372354243, as this is related.

cql/CollateManagementData.cql
```diff
@@ -387,6 +387,27 @@ define Cin2orCin3Biopsies:
...
+
+define MostRecentCin2Biopsy:
+  First(Cin2Biopsies)
+  
+define HsilBiopsies:
+  SortedBiopsyReports S
+    where S.riskTableInput in {'HSIL'}
+
+define MostRecentHsilBiopsy:
+  First(HsilBiopsies)
```
mickohanlon23, line 409: What are these expressions supposed to be doing? Are they looking for the most recent "Hsil, Unspecified" biopsy? There is no `riskTableInput` in the [BiopsyInterpretation function](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/blob/cin2-surveillance/cql/DashboardLibrary.cql#L210-L226) that matches with 'HSIL'.

If that is the case, the name of these expressions should also be changed from Hsil to UnspecifiedHsil.

cql/ManageRareAbnormality.cql
mickohanlon23: Thanks. There are still a couple of expressions added to CollateManagementData.cql that I believe need to be changed from "Hsil" to "HsilUnspecified". I opened https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/118/files#r1383701330 to address this, and will close this comment since the "Hsil" instances have been resolved within this file.

cql/ManageRareAbnormality.cql
mickohanlon23: Names were updated in each of the places mentioned in my comment above, as well as areas where those expressions are referenced. Thanks.

cql/ManageRareAbnormality.cql
mickohanlon23: @mrnosal are you planning to address this? This is the comment I was referring to: https://mitre.sharepoint.com/:w:/r/sites/CervicalCancerCDS/Shared%20Documents/CDS%20Development/L2%20Artifacts/Working%20Documents/CDC_CC_CDS_L2Development_v7.docx?d=wc90d0228d64f420bb1565fb44a712630&csf=1&web=1&e=6ocGpy&nav=eyJjIjo2ODEyNjYyNzh9.

The line that I linked above from the L2 does not seem to be reflected here in the L3. Is there a reason it was omitted, or should it be added?

cql/CollateManagementData.cql
mickohanlon23: I see what you're saying. It appears these functions here:
- Take a subset of SortedBiopsyReports which includes just biopsy reports of a given type (ex. "Cin3Biopsies")
- Take the most recent of all the biopsies that have a result of that type (ex. MostRecentCin3Biopsy)

Whereas there seem to be two primary types of expressions for histologies over in ManageRareAbnormality.cql:
1. "HistologyInterpretedAsCin3" - takes the list of _all_ biopsies, and returns the most recent one if it happens to have a result of CIN3.
2. "AnyHistologyInterpretedAsCin3" - returns a boolean: 'true' if any biopsy result has ever been CIN3, `false` if otherwise.

I see now the distinction, but I still think there could be some room for consolidation here. Some thoughts:
- I think we should consider creating a standard dictionary for how we label these type of lab results, and maybe consider consolidating all of the expressions around test results into the same file, with better commenting to express what is meant by each expression, if the expression name could be interpreted in multiple ways
- Some of the expressions defined here could be used as the base to many of the expressions in ManageRareAbnormality.cql. For example, `AnyHistologyInterpretedAsCin3` is essentially just the boolean version of `MostRecentCin3Biopsy`. `MostRecentCin2BiopsyInPast2Years` is essentially just `MostRecentCin2Biopsy` but with one extra line to institute a 2-year lookback. I building expressions from existing expressions like this could make things more efficient and repeatable and could make it clearer to track down how things are defined.

These things aren't actually affecting the accuracy of the sections modified in this PR, so I would be fine with these things addressed by a follow-up PR around refactoring how we query lab tests, maybe after we discuss some of the thoughts above. @mrnosal If you do plan on leaving this unchanged for this PR, would you mind opening a Jira ticket so that we can track the thoughts I outlined above, as well as any follow-up thoughts you have?

cql/CollateManagementData.cql
mickohanlon23: @mrnosal Any thoughts on this? This comment refers to `HsilBiopsies` and `MostRecentHsilBiopsy`.

cql/ManageRareAbnormality.cql
mrnosal: I'm addressing this in the HSIL branch for PR120

cql/CollateManagementData.cql
mrnosal: I'm addressing this in the HSIL branch for PR120 - I'll change the names of the expressions.

cql/ManageRareAbnormality.cql
mickohanlon23: I really like the refactoring to make this a function! However, I believe the last line of the function needs a `not` before it, since you are checking whether a previous treatment did not exist or if the previous treatment **_was not_** in the time interval of 7 years before the lookback date to now.

CURRENT FUNCTION: 
```cql
define function HadNoCervicalPrecancerTreatmentFromDateToPresent(d System.DateTime, Lookback System.Quantity):
  Collate.DateOfLastCervicalPrecancerTreatment is null or 
  Collate.DateOfLastCervicalPrecancerTreatment in Interval[d - Lookback, Now()]
```

SUGGESTED CHANGE:
```cql
define function HadNoCervicalPrecancerTreatmentFromDateToPresent(d System.DateTime, Lookback System.Quantity):
  Collate.DateOfLastCervicalPrecancerTreatment is null or 
  not (Collate.DateOfLastCervicalPrecancerTreatment in Interval[d - Lookback, Now()])
```

mickohanlon23: Nice! I think this makes these actual recommendation statements a lot cleaner and more readable. And I like the use of the function to make this parametrizable for different lookback periods and for different dates.

cql/ManageRareAbnormality.cql
mickohanlon23: Removed in commit https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/118/commits/c2704894cf2fdb29944fdfe2ff110d3d4094bda7.

cql/ManageRareAbnormality.cql
mickohanlon23: I saw in commit https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/118/commits/c2704894cf2fdb29944fdfe2ff110d3d4094bda7 that the comment was removed, and that you added `and not HasHistologyAfterTreatmentForHighGradeResult`. I still have some questions about the reasoning for adding that line to the L2, but that is more of a question for the clinical team. I documented this question here: [L2 document comment](https://mitre.sharepoint.com/:w:/r/sites/CervicalCancerCDS/Shared%20Documents/CDS%20Development/L2%20Artifacts/Working%20Documents/CDC_CC_CDS_L2Development_v7.docx?d=wc90d0228d64f420bb1565fb44a712630&csf=1&web=1&e=NBNky1&nav=eyJjIjo2ODc4MjMwNjB9). Since this now matches what is in the L2 document, and this PR is not intended to fix section J, I will resolve this comment.

cql/ManageSpecialPopulation.cql
mickohanlon23: Nice! Resolved in https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/118/commits/c2704894cf2fdb29944fdfe2ff110d3d4094bda7. I think this makes things a lot more readable and repeatable.

cql/ManageRareAbnormality.cql
mickohanlon23: Great! This removes the warning that was thrown by trying to `union` two singletons, and I think it also makes the section easier to read.

cql/ManageRareAbnormality.cql
mickohanlon23: This was addressed in PR #120:

https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/120/files#diff-79dcbe5d6961c9c3f6bda93de156d7f281b67a45deecf191662c1ac8cb1c9579.

I will resolve here.

cql/ManageRareAbnormality.cql
mickohanlon23: Mike came up with the labelling of "LastKnown______" to define the pattern here, where we are looking for the most recent biopsy out of the list of biopsies that are specifically resulted as "CIN2". This has been updated throughout PR 120, in commit https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/120/files#diff-79dcbe5d6961c9c3f6bda93de156d7f281b67a45deecf191662c1ac8cb1c9579. I like this approach!

cql/ManageRareAbnormality.cql
mickohanlon23: Discussed this with @mrnosal and we decided on the labelling of "LastKnownX" here to avoid confusion with other expressions in our CQL that are titled "MostRecentX". As for the initial comment about repeating the filtering logic of the biopsies several times in this section, I have opened the following Jira ticket to address this in a follow-on refactor:

**CQL Refactor: Organize and Combine Biopsy Lookbacks**: [CCSMCDS-81](https://jira.mitre.org/browse/CCSMCDS-81)

cql/ManageRareAbnormality.cql
mickohanlon23: Both points 1 and 2 that I brought up here were addressed in the resolved comment above this: https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/118#discussion_r1395066583. I will resolve this thread too, since it touches on the same points.

cql/ManageRareAbnormality.cql
mickohanlon23: Wow I add some redundant comments, sorry about that. This was clarified by the use of "LastKnownX" to distinguish it from other areas where we use "MostRecentX"

cql/CollateManagementData.cql
mickohanlon23: - Names were changed in PR 120 by commit https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/120/commits/d10. This addresses the "standard dictionary" idea brought up here.
- As for consolidating and refactoring, I realize that this may be a bigger undertaking than just what is involved in this PR, so I created a Jira ticket at the bottom of the backlog to address it: [CCSMCDS-81](https://jira.mitre.org/browse/CCSMCDS-81).

cql/ManageRareAbnormality.cql
mrnosal: Thanks for catching this. This was an omission - I started with the function confirming the presence of treatment and then changed it to confirm absence of treatment. I left out the not( ) part.

cql/CollateManagementData.cql
mrnosal: You're right about the 'HSIL, Unspecified'. 798efb6 fixes this.

cql/ManageRareAbnormality.cql
mickohanlon23: No problem. I did some local testing by tweaking an existing test file to check that the new functionality with the added `not` works, but it could be worth adding a test to address this case, since it didn't seem to get caught in any of the existing test cases. The CQL for this is resolved in https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/commit/798efb65d50eeacb95ff027494fe6bebc1d04660, so I'll resolve this comment thread.

mickohanlon23:
Can this be closed since it was incorporated within https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/120, which was merged, @mrnosal?

mrnosal:
Yes, these changes were incorporated into the hsil-unspecified-i1 branch, so this can be closed.

PR #122, opened 2023-12-20
Add 3 year lookback restriction to K.1 Heading 1 for CCSMCDS-44

mrnosal:
The change made by Sharon on Aug 15, 2023 in the L2 added the restriction that the most recent 'under 25 and low-grade cytology result' be <= 3 years ago.
I added this restriction to ONLY K.1 heading 1

### DIFF
--- a/cql/ManageSpecialPopulation.cql
+++ b/cql/ManageSpecialPopulation.cql
@@ -137,6 +137,28 @@ define Under25AndLowGradeCytologyResults:
     )
   )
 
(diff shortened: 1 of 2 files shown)


mickohanlon23 (approved):
Looks good. Tests pass, and this matches the L2 update in Section K.1 Heading 1.

In order to do a rough test that this new lookback was working, I modified the test case for this recommendation so that the cytology results occurred 4 years ago instead of this year. As a result, the test cases failed, as expected, because the cytology results in question were outside of the new lookback.

PR #123, opened 2023-12-20
Change K.1.1 part 3 lookback interval for 2 year visit to 18 months

mrnosal:
The logic for Section K.1 (patients under 25 after abnormal screening result) heading 3 has been updated to look for most recent low grade cytology results 18 months to 3 years after initial low grade cytology result. Previously it was 2 years to 3 years. 
Note that K.1 5a and 5b have not changed their look-back interval (still >= 2 years and <=3 years), so there is additional code to handle both intervals. If the look-back for these two is updated to 18mo, then we can simplify the code. 
This fixes ccsmcds-45

### DIFF
--- a/cql/ManageSpecialPopulation.cql
+++ b/cql/ManageSpecialPopulation.cql
@@ -89,12 +89,6 @@ define Associated18MonthHpvCotest:
       where H.date within 1 day of Under25And18MonthsAgoCytologyReport.date
   )
 
-define Associated2YearHpvCotest:
-  Coalesce(
-    (Collate.SortedHpvReports) H
-      where H.date within 1 day of Under25And2YearsAgoCytologyReport.date
-  )
-
 define HistologicHsilCin2OrUnspecified:
   First(
     (Collate.SortedBiopsyReports) B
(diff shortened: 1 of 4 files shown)


mrnosal:
GM agrees that K.1 5a and 5b should also be changed to >= 18mo and <=3 years to match K.1 heading 3

mickohanlon23 (approved):
Generally this update looks good and aligns with the L2 for this section. However, as you pointed out by the [comment in the test case you added](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/123/files#diff-1b402d66b8ba911414775bbf048be841e81576bde3be5744b3e4c2e456018f18R4), this update changes K.1, Heading 1, Recommendation C (the 3rd recommendation in K.1, Heading 1, rather than K.1 Heading 3, as you describe in the PR description and as is described in [CCSMCDS-45](https://jira.mitre.org/browse/CCSMCDS-45). Therefore, this update appears to address one of the surveillance interval updated covered by [CCSMCDS-72](https://jira.mitre.org/browse/CCSMCDS-72), but does not actually address the change request in K.1 Heading 3 denoted by [CCSMCDS-45](https://jira.mitre.org/browse/CCSMCDS-45). If you search for "Heading 3" in ManageSpecialPopulation.cql, you will see where this code is.

mickohanlon23:
I approved since this code is correct and contributes to [CCSMCDS-72](https://jira.mitre.org/browse/CCSMCDS-72), but a new pull request should be opened to address [CCSMCDS-45](https://jira.mitre.org/browse/CCSMCDS-45).

As for this branch, we can either merge as-is, or add the rest of the surveillance interval updates to this branch. Either way, I would propose changing the name and description of this pull request and of the ultimate squash commit name to avoid confusion.

PR #130, opened 2024-01-20
Modify I.6 to support care when AIS is found during an excision

mickohanlon23:
This PR addresses Jira [CCSMCDS-60](https://jira.mitre.org/browse/CCSMCDS-60).

Section I.6 had logic for managing AIS that was followed up by a cervical excision. However, there was not logic to support AIS that was found _from_ a cervical excision. This logic expands two of the I.6 recommendations to not only make a recommendation when treatment is found after AIS, but also when treatment is found _on the same date_ as AIS.

This should work with our pap tracker translation code in https://github.com/ccsm-cds-tools/ccsm-cds-service/blob/pilot/config/apply/ccsm/translate.js, since a cervical excision that was resulted as AIS in the pap tracker gets translated to both a cervix excision procedure, as well as an AIS biopsy result.


mrnosal (approved):
Code reviewed and test-cql run.

PR #132, opened 2024-01-23
Update Observation Logic Section K.1 Heading 3

mrnosal:
This addresses the latest L2 observation logic for Section K.1 Heading 3 - Management of Histology of Less than CIN2 Preceded by Cytology ASC-H and HSIL in Patients Younger than 25 Year

The logic was updated to check for cytology or histology after 'CIN1 or less preceded by ASC-H/HSIL' by looking for a count of < 2. This means there are two possible cases: the cytology or histology result is present or there are zero results present. Adding a COUNT clause to the CQL logic complicates looking for the original CIN1 or less result.
Therefore, to simplify the logic, these checks were added as individual cases (look for zero results after the CIN1 or less, look for one result after the CIN1 or less). This adds some redundancy to the code, but makes inspection of the code easier. 

The time range for the "persists for 2 years" is specified as ">2 years" so test results which were *exactly* two years apart in the test cases need to be adjusted. 

Additional logic diagrams have been linked to the JIRA ticket CCSMCDS-45 for this PR.

### DIFF
--- a/cql/CollateManagementData.cql
+++ b/cql/CollateManagementData.cql
@@ -438,8 +438,14 @@ define AisBiopsies:
   SortedBiopsyReports S
     where S.riskTableInput in {'AIS'}
 
+// I believe this is mis-named - it does not refer to the Most Recent Biopsy being AIS
+// but instead finds the chronologically newest of all AIS biopsies
+// This is what should be called LastKnownAisBiopsy
 define MostRecentAisBiopsy:
   First(AisBiopsies)
+  
+define LastKnownAisBiopsy:
+  First(AisBiopsies) 
   
 // Look for treatments after this CIN2/CIN3 biopsy
 define CervicalPrecancerTreatments:
--- a/cql/ManageSpecialPopulation.cql
+++ b/cql/ManageSpecialPopulation.cql
@@ -244,8 +244,15 @@ define RecommendationForPatientsYoungerThan25:
         Rare.CytologyInterpretedAsHsil or
         Rare.CytologyInterpretedAsAscH
       ) and
-      Rare.MostRecentCytologyReport.date more than 18 months after Collate.MostRecentBiopsyReport.date and
-      Rare.MostRecentCytologyReport.date less than 3 years after Collate.MostRecentBiopsyReport.date
+      Rare.MostRecentCytologyReport.date more than 2 years after Collate.MostRecentBiopsyReport.date and
+      Rare.MostRecentCytologyReport.date less than 3 years after Collate.MostRecentBiopsyReport.date and
+      not (
+      // cannot find these in the 2-3 year window after the Histology interpreted as Cin1OrNormal
+      // NOT histologic HSIL (CIN2)
+      // NOT Histologic HSIL, Unspecified
+      // NOT Histologic AIS
+      // NOT Histologic Cancer
+      )
     ) then
       {
         short: 'See Details',
--- a/cql/ManageSpecialPopulation.cql
+++ b/cql/ManageSpecialPopulation.cql
@@ -217,10 +217,20 @@ define HistologyInterpretedAsLessThanCin2AfterAbnormalCytologyScreening:
   Collate.MostRecentBiopsyReport.date 3 years or less after Rare.MostRecentCytologyReport.date
 
 define SecondMostRecentCytologyInterpretedAsAscHOrHsil:
+  SecondMostRecentCytologyInterpretedAsAscH or
+  SecondMostRecentCytologyInterpretedAsHsil
+
+define SecondMostRecentCytologyInterpretedAsAscH:
   AnyTrue(
     (Rare.SecondMostRecentCytologyReport.allConclusions) aC
       return
-        aC ~ "ASC-H" or
+        aC ~ "ASC-H"
+  )
+
+define SecondMostRecentCytologyInterpretedAsHsil:
+  AnyTrue(
+    (Rare.SecondMostRecentCytologyReport.allConclusions) aC
+      return
         aC in "HSIL"
   )
 
@@ -245,14 +255,14 @@ define RecommendationForPatientsYoungerThan25:
         Rare.CytologyInterpretedAsAscH
       ) and
       Rare.MostRecentCytologyReport.date more than 2 years after Collate.MostRecentBiopsyReport.date and
-      Rare.MostRecentCytologyReport.date less than 3 years after Collate.MostRecentBiopsyReport.date and
-      not (
+      Rare.MostRecentCytologyReport.date less than 3 years after Collate.MostRecentBiopsyReport.date
+      // and    not (
       // cannot find these in the 2-3 year window after the Histology interpreted as Cin1OrNormal
       // NOT histologic HSIL (CIN2)
       // NOT Histologic HSIL, Unspecified
       // NOT Histologic AIS
       // NOT Histologic Cancer
-      )
+//      )
     ) then
       {
         short: 'See Details',
@@ -263,8 +273,9 @@ define RecommendationForPatientsYoungerThan25:
           'A diagnostic excisional procedure is recommended in patients when the squamocolumnar junction or the upper limit of all lesions are not fully visualized.'
         }
       }    
+
     when ( // Heading 3.2
-      AgeInYears() >= 21 and AgeInYears() < 25 and
+      AgeInYears() >= 21 and AgeInYears() < 25 and 
       Rare.HistologyInterpretedAsCin1OrNormal and
       (
         Collate.CytologyInterpretedAsAscusOrAbove or
@@ -282,11 +293,17 @@ define RecommendationForPatientsYoungerThan25:
           'Colposcopy is recommended.'
         }
       }    
-    when ( // Heading 3.1b
+
+// ===========    
+// K.1 Heading 3 Heading Part 1
+
+    when ( // Heading K.1.3.1 case a
       AgeInYears() < 25 and
       Rare.HistologyInterpretedAsCin1OrNormal and
-      Rare.CytologyInterpretedAsHsil and
-      Rare.MostRecentCytologyReport.date less than 12 months before Collate.MostRecentBiopsyReport.date
+      Rare.CytologyInterpretedAsHsil
+      // How far back can the CIN1 Histology report be?
+      // Rare.MostRecentCytologyReport.date less than 12 months before Collate.MostRecentBiopsyReport.date
+      // DO WE HAVE A DATE RANGE FOR THIS HSIL BEFORE THE CIN1?
     ) then
       {
         short: 'Colposcopy and Cytology',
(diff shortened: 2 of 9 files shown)


cql/ManageSpecialPopulation.cql
```diff
@@ -217,10 +217,20 @@ define HistologyInterpretedAsLessThanCin2AfterAbnormalCytologyScreening:
...
+        aC ~ "ASC-H"
+  )
+
+define SecondMostRecentCytologyInterpretedAsHsil:
   AnyTrue(
     (Rare.SecondMostRecentCytologyReport.allConclusions) aC
       return
-        aC ~ "ASC-H" or
         aC in "HSIL"
   )
```
mickohanlon23, line 235: `SecondMostRecentCytologyInterpretedAsAscHOrHsil`,`SecondMostRecentCytologyInterpretedAsAscH`,`SecondMostRecentCytologyInterpretedAsHsil`:

Should we also put these 3 expressions over in ManageRareAbnormality.cql, with `CytologyInterpretedAsAscHOrHsil`, which you added there?

cql/ManageSpecialPopulation.cql
```diff
@@ -256,16 +268,88 @@ define RecommendationForPatientsYoungerThan25:
           'A diagnostic excisional procedure is recommended in patients when the squamocolumnar junction or the upper limit of all lesions are not fully visualized.'
         }
       }    
-    when ( // Heading 3.2
-      AgeInYears() >= 21 and AgeInYears() < 25 and
+
+    when ( // K.1 Heading 3 case 2
```
mickohanlon23, line 272: The recommendation for CIN1 preceded by HSIL is to get a colpo and cytology in 1 year. The recommendation for CIN1 preceded by ASC-H is to get a cytology in 1 year. If a patient followed the guidelines, in order to have one of these high grade cytology results persist for two years, they would have had to have had the original CIN1 preceded by ASC-H/HSIL, followed by another cytology or cytology+colpo at the 1 year mark, followed by yet a 3rd cytology+colpo at the 2 year mark. Therefore, I believe labelling the original CIN1 preceded by HSIL/ASCH here as the "SecondMostRecent" breaks down. 

The test you added - `Young33HSILCytoCin1Persists.yml` - is not representative of what somebody following the guidelines would do, because the synthetic patient waited over 2 years to come back for a follow up, instead of coming back in 1 year.

cql/ManageSpecialPopulation.cql
```diff
@@ -236,16 +246,18 @@ define ShouldSwitchToRiskAt25Text:
 define RecommendationForPatientsYoungerThan25:
   case
     // Heading 3: Management of Histology of Less than CIN2 Preceded by Cytology ASC-H and HSIL in Patients Younger than 25 Year
-    when ( // Heading 3.3
+    // ===========    
+    when ( // K.1 Heading 3 case 1
```
mickohanlon23, line 250: Same sentiment as comment below but for this rec - original cytology would likely not be the 2nd most recent, for reason explained blow.

cql/ManageSpecialPopulation.cql
mickohanlon23: One idea of how to frame this logic would be to look for the most recent cytology to be ASC-H or HSIL, and then if that's true, check if there is ANY CIN1 preceded by ASC-H or HSIL cytology between 2-3 years before that most recent cytology test.

cql/ManageSpecialPopulation.cql
```diff
@@ -274,8 +358,12 @@ define RecommendationForPatientsYoungerThan25:
         details: {
           'Colposcopy is recommended.'
         }
-      }    
-    when ( // Heading 3.1b
+      }
+
+// ===========    
+// K.1 Heading 3 Heading Part 1
```
mickohanlon23, line 364: I don't think we need to left indent all the way for a subheader. This can probably be indented the same amount as the `when` below it, and then written without the spaces. After all, this is a sub-sub-sub header (or something like that), so I don't think we want to make the break as dramatic as between the higher-level headers.

cql/ManageSpecialPopulation.cql
```diff
@@ -291,18 +379,61 @@ define RecommendationForPatientsYoungerThan25:
           '*Refer to ASCCP Consensus Guideline 2019 Section K.1'
         }
       }
-    when ( // Heading 3.1c
+
+    when ( // Heading K.1.3.1 case b
```
mickohanlon23, line 383: Styling suggestion: If these 3 cases are for the same recommendation in the L2, then I think it may be more consistent to combine them in one "when" clause that contains 3 "OR" portions. You could include comments within each of the "OR" portions to denote what each part is looking for. In my opinion, that might make it clearer that each of the recommendations are actually different parts of the same recommendation in the L2.

I know you separated them on purpose to facilitate inspection of the code, so open to differing opinions if you think these recs are more readable as-is.

cql/ManageSpecialPopulation.cql
```diff
@@ -291,18 +379,61 @@ define RecommendationForPatientsYoungerThan25:
...
+        }
+      }
+
+    when ( // Heading K.1.3.1 case d
+      // 1 cyto and 1 biopsy after CIN1 preceded by HSIL
+      AgeInYears() < 25 and
+      Rare.SecondMostRecentHistologyInterpretedAsCin1OrNormal and
+      SecondMostRecentCytologyInterpretedAsHsil and
+      Rare.SecondMostRecentCytologyReport.date less than 12 months before Rare.SecondMostRecentBiopsyReport.date and
+      Collate.MostRecentBiopsyReport.date less than 3 years after Rare.SecondMostRecentBiopsyReport.date and
```
mickohanlon23, line 428:
```suggestion
      Rare.SecondMostRecentCytologyReport.date less than 12 months before Rare.SecondMostRecentBiopsyReport.date and
      Rare.HistologyInterpretedAsCin1OrNormal and
      Rare.CytologyInterpretedAsNilm and
      Collate.MostRecentBiopsyReport.date less than 3 years after Rare.SecondMostRecentBiopsyReport.date and
```

Don't we need to check for the results of the most recent tests?

test/ManagementSpecialPopulations/cases/Young33HSILCytoCin1Persists.yml
```diff
@@ -0,0 +1,34 @@
+---
+name: Younger Than 25 Cytologic HSIL or ASC-H With Histologic CIN1 Persists for 2 years 3.3
```
mickohanlon23, line 2: This test is not indicative of what a patient following the guidelines would have done. They also would have had a follow up cytology+colpo (or just cytology for ASC-H) at the 1 year mark, i.e. 1 year ago.

cql/ManageSpecialPopulation.cql
mickohanlon23, line 250: Should we designate that these are K.1, Heading 3, part 3, and label as K.1.3.3, to align with how you have labelled part 1 and part 2?

mickohanlon23 (changes requested):
This works for the most part, except I think there may be flaws in K.1 Heading 3 recommendation 3 (persistent ASC-H / HSIL). There was one other error I pointed out, as well as a few styling suggestions. Let me know what you think.

cql/CollateManagementData.cql
```diff
@@ -438,9 +438,15 @@ define AisBiopsies:
   SortedBiopsyReports S
     where S.riskTableInput in {'AIS'}
 
+// I believe this is mis-named - it does not refer to the Most Recent Biopsy being AIS
+// but instead finds the chronologically newest of all AIS biopsies
+// This is what should be called LastKnownAisBiopsy
```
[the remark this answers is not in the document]
mrnosal: I ended up doing this as part of fixing ccsmcds-107. It is in [PR 134](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/134)
I'll remove the comment here.

cql/ManageSpecialPopulation.cql
mrnosal: I agree it makes sense to put these in ManageRareAbnormality. Will move.

cql/ManageSpecialPopulation.cql
mrnosal: I believe there is overlap in the L2 logic for K.1.3.2(b) and K.1.3.3. We should discuss how to resolve.

cql/ManageSpecialPopulation.cql
mickohanlon23: I still think these make more sense lower down in the ManageRareAbnormality.cql file, right after where you added `CytologyInterpretedAsAscHOrHsil`, since these expressions are very closely related. They don't really play a part in the Rare Cytology section where you moved them to, is there a reason they were added there? I won't let this hold up the PR, since functionally it doesn't change a thing but I leave it up to you if you would like to change it.

cql/ManageSpecialPopulation.cql
mickohanlon23: You're right, there is. That's the reason we evaluate K.1.3.3 before K.1.3.2(b). Because the recommendation in K.1.3.3 is more specific. But that's an entirely different issue.

What I was describing above is that K.1.3.3 is a recommendation for "If a high-grade cytologic abnormality (HSIL, ASC-H) without histologic HSIL persists for 2 years". It's not saying that if your most recent and second most recent test were HSIL/ASC-H, than you should get a diagnostic excisional procedure. It's saying that 2-3 years ago they had HSIL/ASC-H, and now they still do. But those two tests don't have to be subsequent, there can, and should, be other tests in between. You've inserted "MostRecent" and "SecondMostRecent" logic here that doesn't align with recommendation K.1.3.3 in the L2.

cql/ManageSpecialPopulation.cql
mickohanlon23: I saw you updated this in your most recent commit. I think this is a lot easier to track now against the L2 and against the guidelines!

cql/ManageSpecialPopulation.cql
mickohanlon23: It looks like this still needs to be addressed.

test/ManagementSpecialPopulations/cases/Young33HSILCytoCin1Persists.yml
mickohanlon23: Still needs to be addressed. This is related to https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/132#discussion_r1465485738

cql/ManageSpecialPopulation.cql
mickohanlon23: This makes things a lot clearer, thanks!

cql/ManageSpecialPopulation.cql
mrnosal: These are no longer needed due to refactoring. They will be removed.

cql/ManageSpecialPopulation.cql
mrnosal: Removed the use of SecondMostRecent here. Changed logic to check if there is ANY CIN1 preceded by ASC-H or HSIL cytology between 2-3 years before that most recent cytology test AND check there is no treatment after the most recent cytology.

cql/ManageSpecialPopulation.cql
mrnosal: concur - switched to use PrecedingCytologyIsHsil and PrecedingCytologyIsAscH instead.

cql/ManageSpecialPopulation.cql
mrnosal: Now checks most recent Histology and cytology in 3 year window after most recent or second most recent biopsy report.

test/ManagementSpecialPopulations/cases/Young33HSILCytoCin1Persists.yml
mrnosal: This test case is not indicative of following guidelines, but patient who missed 1 year follow up and returned with persistent ASC-H or HSIL would still get this recommendation. Adding a new test case to include a 'normal' 1 year followup biopsy+cytology.

mrnosal:
Refactored section K.1 Heading 3 code. Parts 1 and 2 were refactored to use PrecedingCytologyIsAscH and PrecedingCytologyIsHSIL instead of relying on SecondMostRecentCytology. K.1 Heading 3 part 3 was significantly refactored to use approach discussed. (Option B: Looking for “Most Recent” CytologyInterpretedAsAscH and checking that the patient hasn’t gotten a treatment for it yet.) Added extra test case for patient receiving normal followup at 1 year, but still ASC-H/HSIL persists at 2 year followup.

PR #133, opened 2024-02-12
Update Logic for J.2.1, J.2.3, J.3.1, remove J.2.2

mrnosal:
Update the Logic in Section J.2.1, J2.3 and J.3.1. Section J.2.2 is not needed as the patient should be managed by Common Abnormality Table 5A.
J.2.1 - add checks to make sure patient does not have surveillance, cytology alone, colpo or histology after the most recent treatment.
J.2.3 - update HSIL lookback to <=7 years after treatment
J.3.1 Update treatment after high grade cytology to 18mo lookback, add check for no histology, cytology aSC-US or above, or HPV+ after the treatment.


mickohanlon23:
@mrnosal I just saw your Jira comment that the pilot patients were tested and are now getting correct recommendations. You can go ahead and merge this after resolving conflicts with main branch. Thanks!

PR #139, opened 2024-05-21
Add Support for DiagnosticReport Lookback date

mickohanlon23:
This PR addresses the logic changes needed in this repo to handle [CCSMCDS-157](https://jira.mitre.org/browse/CCSMCDS-157).

This PR:
- Overloads the `Lookback()` function for Diagnostic Reports, so that a `LookBackDate` can be passed in to get only DiagnosticReports within the lookback period that _occur after a certain date_.
- Adds CQL Parameters in Dashboard.cql for `HrHPVLookbackDate` and `CervicalCytologyLookbackDate`
- Passes new LookBackDate parameters into CQL expressions that aggregate cervical cytology tests and HrHPV tests.
- Bumps the package version to 0.4.17


mrnosal (approved):
LookBack with date works as expected. If LookBackDate is within the LookBack interval, it correctly restricts matching reports. If LBD is > Now(), then null results returned. Tests pass.

PR #140, opened 2024-06-28
Update recommendation text relative to latest test

mrnosal:
Update recommendation details when the time interval mentioned in the text is relative to the date of the last test (e.g. most recent cytology, histology, etc.) 

It can be confusing if the recommendation says "1 one year" of a test which occurred 15 months in the past. 
This change adopts the phrase "date of most recent test" to be the reference the time interval applies to. The date value shown is not changed (e.g. Collate.DateOfMostRecentReport + 1 years) but will now make it clear in the recommendation details that it is based on the most recent test, and that's why the date shown may be past.

### DIFF
--- a/cql/ManageCommonAbnormality.cql
+++ b/cql/ManageCommonAbnormality.cql
@@ -491,7 +491,7 @@ define ErrataRecommendation:
         short: 'Reflex HPV',
         date: Today(),
         details: {
-          'Reflex HPV testing is preferred. If HPV testing is not feasible, repeat cytology within 1 year from date of most recent test.'
+          'Reflex HPV testing is preferred. If HPV testing is not feasible, repeat cytology in 1 year from date of most recent test.'
         }
       }
     when ( // Errata Recommendation 3.1.2: LSIL Alone
@@ -571,7 +571,7 @@ define ErrataRecommendation:
         short: 'Surveillance',
         date: Collate.DateOfMostRecentReport + 1 years,
         details: {
-          'Primary HPV testing or cotesting is recommended within one year from date of most recent test.'
+          'Primary HPV testing or cotesting is recommended in 1 year from date of most recent test.'
         }
       }
     else
@@ -594,13 +594,13 @@ define TreatmentWithPregnancyConcernsRecommendationText:
   'If the patient is currently pregnant, endocervical curettage, endometrial biopsy, and treatment without biopsy are unacceptable.'
 
 define OneYearSurveillanceRecommendationText:
-  'Primary HPV testing or cotesting is recommended within 1 year of most recent test.'
+  'Primary HPV testing or cotesting is recommended in 1 year from date of most recent test.'
   
 define ThreeYearSurveillanceRecommendationText:
-  'Primary HPV testing or cotesting is recommended within 3 years of most recent test.'
+  'Primary HPV testing or cotesting is recommended in 3 years from date of most recent test.'
   
 define FiveYearSurveillanceRecommendationText:
-  'Primary HPV testing or cotesting is recommended within 5 years of most recent test.'
+  'Primary HPV testing or cotesting is recommended in 5 years from date of most recent test.'
 
 define FiveYearSurveillanceIntervalText:
   'Future testing is recommended at 5-year intervals.'  
--- a/cql/ManageCommonAbnormality.cql
+++ b/cql/ManageCommonAbnormality.cql
@@ -571,7 +571,7 @@ define ErrataRecommendation:
         short: 'Surveillance',
         date: Collate.DateOfMostRecentReport + 1 years,
         details: {
-          'Primary HPV testing or cotesting is recommended in one year.'
+          'Primary HPV testing or cotesting is recommended within one year from date of most recent test.'
         }
       }
     else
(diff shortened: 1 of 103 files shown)


cql/ManageRareAbnormality.cql
```diff
@@ -402,7 +402,7 @@ define RecommendationForRareCytology:
         date: Collate.DateOfMostRecentReport + 1 year,
```
mickohanlon23, line 402: the following recommendations in `ManageRareAbnormality.cql` could also benefit from this same `from date of most recent test` update:
- I3.3
- I4.3 - Recommendation # 1
- I4.1
- I4.2

cql/ManageCommonAbnormality.cql
```diff
@@ -571,7 +571,7 @@ define ErrataRecommendation:
         short: 'Surveillance',
```
mickohanlon23, line 571: `Errata Recommendation 3.1.1: ASC-US Alone` in ManageCommonAbnormality.cql could benefit from this same style of update.

cql/ManageSpecialPopulation.cql
```diff
@@ -302,7 +302,7 @@ define RecommendationForPatientsYoungerThan25:
         date: Collate.DateOfMostRecentReport + 1 year,
```
mickohanlon23, line 302: Heading 3.1b in `ManageSpecialPopulation.cql` could also benefit from an update.

mickohanlon23 (changes requested):
Thanks for making this update. I added some inline comments about two recommendations, and then suggestions for other recommendations that I think could benefit from an upgrade. I have two main high-level questions:
1. Has this update been made in L2 yet? Before merging this, we should make sure the L2 is up-to-date with the changes we are making here.
2. What was your method for choosing which recommendation statements to update? Did you have a list based upon user feedback? By searching for things like "1 year" and "Surveillance" in the files that contain Management recommendations, I found several cases that follow the same recommendation pattern of a future follow-up test in X years that you did not update. I think it would be beneficial to do a comprehensive walkthrough of our recommendations before merging this PR, so that we are not adding inconsistency to our recommendations by adopting the "from date of most recent test" language in some surveillance recommendations but not in others.

cql/ManageCommonAbnormality.cql
```diff
@@ -582,25 +582,25 @@ define ReflexTestingRecommendationText:
...
-
+  'Primary HPV testing or cotesting is recommended within 1 year of most recent test.'
+  
 define ThreeYearSurveillanceRecommendationText:
-  'Repeat testing in 3 years with primary HPV testing or cotesting is recommended.'
+  'Primary HPV testing or cotesting is recommended within 3 years of most recent test.'
   
 define FiveYearSurveillanceRecommendationText:
-  'Routine screening at 5-year intervals using primary HPV testing or cotesting is recommended.'
+  'Primary HPV testing or cotesting is recommended at 5-year intervals from date of most recent test.'
```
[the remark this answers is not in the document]
mrnosal: Will adjust text to separate the next upcoming test and future intervals.

cql/ManageSpecialPopulation.cql
```diff
@@ -728,7 +728,7 @@ define RecommendationForManagingPatientsAfterHysterectomy:
         date: Collate.DateOfMostRecentReport + 3 years,
         group: 'Hysterectomy (K.4)',
         details: {
-          'Primary hrHPV testing or cotesting is recommended every 3 years for 25 years after hysterectomy is performed for treatment of histologic HSIL, CIN 2, CIN 3 or AIS.'
+          'Primary hrHPV testing or cotesting is recommended every 3 years after date of most recent test for 25 years after hysterectomy is performed for treatment of histologic HSIL, CIN 2, CIN 3 or AIS.'
```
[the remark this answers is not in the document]
mrnosal: Will adjust to separate the next test, the 25-year period and the 3-year intervals to make it clear.

cql/ManageRareAbnormality.cql
mrnosal: Update I 3.3 to "1 year from date of most recent test"
Update I 4.2 to "1 year after the colposcopy" as per L2.
Update I 4.1 to refer to "colposcopy within 1 year of most recent histology"
Update 4.3 rec 1 to "within 1 year of date of most recent test"

cql/ManageCommonAbnormality.cql
mrnosal: will update to "within 1 year from date of most recent test"

cql/ManageCommonAbnormality.cql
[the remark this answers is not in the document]
mrnosal: Language now reads:
  Primary HPV testing or cotesting is recommended within 5 years of most recent test.
  Future testing is recommended at 5-year intervals.

This keeps the text consistent with the 1-year and 3-year surveillance recommendation text and makes clear the 5-year interval after.

mrnosal:
> 2\. What was your method for choosing which recommendation statements to update?

Statements which included a date offset from some test - e.g. Collate.DateOfMostRecentReport + 1 year
As you noticed, there are some statements which have more than one actionable item in their recommendation, which is also based relative to a test.

cql/ManageSpecialPopulation.cql
mrnosal: Missed that one. Now refers to "within 1 year from date of most recent test"

cql/ManageSpecialPopulation.cql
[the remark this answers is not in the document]
mrnosal: text now reads
          'Primary hrHPV testing or cotesting is recommended within 3 years of most recent test.',
          'Testing should continue for 25 years after hysterectomy is performed for treatment of histologic HSIL, CIN 2, CIN 3 or AIS.',
          'Future testing is recommended at 3-year intervals.'

This keeps actionable item in first line, makes clear the 25 years refers to the interval after the hysterectomy and puts the interval in its own line, like the FiveYearSurveillanceRecommendationText in ManageCommonAbnormality.

cql/ManageCommonAbnormality.cql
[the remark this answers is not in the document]
mickohanlon23: I'm not sure if there is a reason to separate `FiveYearSurveillanceRecommendationText` and `FiveYearSurveillanceIntervalText`. Will these ever be referenced independently from one another?

That would be my one remaining comment in this thread. Overall the updated language sounds a lot more clear, so this is not a critical change, since it doesn't affect what is visible to the user.

cql/ManageRareAbnormality.cql
mickohanlon23: ### Changing "in" to "within":

Where did the "_within_" wording come from? This seems to apply that they can return for testing at any point within the next year, whereas the recommendation is supposed to be that they should not come back for a year. Thoughts?

cql/ManageSpecialPopulation.cql
mickohanlon23: Same comment as I had above, about the fact that changing "in" to "within" slightly changes the meaning of the recommendation. I will keep this thread resolved, so as not to be duplicative, and will keep the other one open for us to discuss there.

cql/ManageSpecialPopulation.cql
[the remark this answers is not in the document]
mickohanlon23: I think this new formal is very clear. Thanks for updating.

mickohanlon23 (changes requested):
Thanks for making the updates. Overall I think we should be in good shape to go ahead and merge this in today, but I had two remaining comments I was hoping you could take a look at before we do so, which is why I marked this as "Request changes" again. Let me know what you think.

cql/ManageCommonAbnormality.cql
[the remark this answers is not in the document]
mickohanlon23: @mrnosal and I discussed over a call, and decided it is fine to keep these recommendation text components separated, as they are in this PR.

cql/ManageRareAbnormality.cql
mrnosal: Have updated instances of "within" to "in" as per disussion. This reflects the guideline's preferred use of "in 1 year", "in 3 years" , etc.

mickohanlon23: Nice! I think changing "within" to "in" clarifies the recommendation text throughout the logic, so that it aligns more closely with the guidelines.

I found one recommendation text that was missed, in Section K3.2, so I added commit https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/140/commits/b68ec596e75474ba4f2791bee70a8f7fdbf1bf6f to change "within" to "in" for this recommendation, as well as update the test case associated with K3.2.

mickohanlon23 (approved):
Looks good! Thanks for making these changes and working through the revisions. As I mentioned in the last comment I resolved, I added a commit to change one final recommendation text instance of "within" to "in" that I found in section K3.2. I made this change in the L2 as well. I also added an additional commit to bump the package version to 0.4.18, so that we can publish a new version with this change to npm.

PR #141, opened 2024-08-15
Specify time intervals in surveillance recommendation short texts

mickohanlon23:
Addresses [CCSMCDS-204](https://jira.mitre.org/browse/CCSMCDS-204). It was requested that our "short" recommendation text statements of "Surveillance" be updated to specify the date at which a patient is due back for testing, since the term surveillance may be confusing to a clinician in this context.

### DIFF
--- a/cql/ManageCommonAbnormality.cql
+++ b/cql/ManageCommonAbnormality.cql
@@ -568,7 +568,7 @@ define ErrataRecommendation:
       )
     ) then
       {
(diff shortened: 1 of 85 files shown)


mrnosal (approved):
These changes make sense for the use of Surveillance to mean cotesting. The time interval is clear in the the short, and explained in the details (1-year from date of most recent rest).

PR #142, opened 2024-08-15
Update 'See Details' short text with more a more specific title for each recommendation

mickohanlon23:
Addresses [CCSMCDS-145](https://jira.mitre.org/browse/CCSMCDS-145). Updates instances where "short" title text was previously "See Details".

I would appreciate a review of these updates, as there is not always one obvious choice for condensing the recommendation into a short expression. Open to feedback regarding wording decisions I made.

### DIFF
--- a/cql/ManageRareAbnormality.cql
+++ b/cql/ManageRareAbnormality.cql
@@ -423,7 +423,7 @@ define RecommendationForRareCytology:
       )
     ) then
       {
-        short: 'See Details',
+        short: 'Diagnostic Excisional Procedure',
         date: Today(),
         group: 'Rare Cytology (G.1.5)',
         details: {
@@ -440,7 +440,7 @@ define RecommendationForRareCytology:
       HasNoCytologyInterpretedAsHsilAfterDate(LastKnownCytologyInterpretedAsAgcOrAisWithoutAtypicalEndometrialCells.date)
     ) then
       {
-        short: 'See Details',
+        short: 'Colposcopy',
         date: Today(),
         group: 'Rare Cytology (G.1.1)',
         details: {
@@ -457,7 +457,7 @@ define RecommendationForRareCytology:
       HasNoCytologyInterpretedAsHsilAfterDate(LastKnownCytologyInterpretedAsAgcOrAisWithoutAtypicalEndometrialCells.date)
     ) then
       {
-        short: 'See Details',
+        short: 'Colposcopy',
         date: Today(),
         group: 'Rare Cytology (G.1.2)',
         details: {
@@ -473,7 +473,7 @@ define RecommendationForRareCytology:
       HasNoCytologyInterpretedAsHsilAfterDate(LastKnownCytologyInterpretedAsAgcOrAisWithAtypicalEndometrialCells.date)
     ) then
       {
-        short: 'See Details',
+        short: 'Endometrial and Endocervical Sampling',
         date: Today(),
         group: 'Rare Cytology (G.1.3)',
         details: {
@@ -535,7 +535,7 @@ define RecommendationForRareCytology:
       CytologyInterpretedAsNilm
     ) then
       {
-        short: 'See Details',
+        short: 'Routine Screening',
         date: Today(),
         group: 'Rare Cytology (G.3.1)',
         details: {
@@ -553,7 +553,7 @@ define RecommendationForRareCytology:
       )
     ) then
       {
-        short: 'See Details',
+        short: 'HPV Testing',
         date: Today(),
         group: 'Rare Cytology (G.3.2)',
         details: {
@@ -574,7 +574,7 @@ define RecommendationForRareCytology:
       )
     ) then
       {
-        short: 'See Details',
+        short: 'No Further Evaluation',
         date: Today(),
         group: 'Rare Cytology (G.4.1)',
         details: {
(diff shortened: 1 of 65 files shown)


cql/ManageRareAbnormality.cql
```diff
@@ -1627,7 +1627,7 @@ define RecommendationForHistologyResults:
       )
     ) then
       {
-        short: 'See Details',
+        short: 'Surveillance',
         date: Today(),
         group: 'Managing Histology (I.5)',
         details: {
```
mrnosal, line 1633: Here the short is 'Surveillance' but details says 'observation' - no mention of HPV testing or cotesting, which is how Surveillance is used elsewhere. Should details be changed to 'surveillance with primary HPV testing or cotesting' ?

cql/ManageRareAbnormality.cql
```diff
@@ -1992,7 +1992,7 @@ define RecommendationForSurveillanceAfterAbnormalities:
       InitialIntensiveSurveillancePeriod
     ) then
       {
-        short: 'See Details',
+        short: 'Surveillance 3-year follow-up',
         date: Today(),
         group: 'Management Surveillance (J.3.2)',
         details: {
```
mrnosal, line 1998: Should details say "Surveillance with primary HPV testing or cotesting" to match other recommendations (e.g. I.3)

cql/ManageSpecialPopulation.cql
```diff
@@ -365,7 +365,7 @@ define RecommendationForPatientsYoungerThan25:
       Collate.HasNoTreatmentAfterDate(Collate.LastKnownCin2Biopsy.date)
     ) then
       {
-        short: 'See Details',
+        short: 'Surveillance 6-month follow-up',
         date: Today(),
         group: 'Younger Than 25 (K.1.4)',
         details: {
```
mrnosal, line 371: Here Surveillance does not mean "HPV or cotesting" as elsewhere. Should review the recommendation text, and see if details needs updating. Should short be changed to something more specific in this case?

cql/ManageSpecialPopulation.cql
```diff
@@ -737,7 +737,7 @@ define RecommendationForManagingPatientsAfterHysterectomy:
       Rare.AbsenceOrRemovalOfCervix
     ) then 
       {
-        short: 'See Details',
+        short: 'Surveillance',
```
mrnosal, line 740: Should this include the '3 year' interval as part of the short text to match other Surveillance recommendations?

cql/ManageRareAbnormality.cql
mickohanlon23: My thought was that most `short` recommendations that say just "Surveillance" imply that the surveillance is with primary HPV testing and cotesting. There are other recommendations with this short text that mention HPV testing or cotesting in the `details`, but not in the `short`. If I changed this here, then I think we would want to change it elsewhere too for consistency.

If the issue is with the fact that "Observation" wasn't mentioned in the `short` text, I was under the assumption that Observation and Surveillance are somewhat synonymous, because we don't have other `short` texts that say 'Observation'. But that could be something we ask our pilot partner.

mickohanlon23: Oh sorry, you were proposing we change the _details_ text, I misread your comment originally. I didn't intend to make any changes to the actual details text in this PR, since I figured that would require a more thorough clinical review, since the detailed text is abstracted directly from the guidelines. I am under the impression that "Observation" here implies that it would be done with primary HPV testing or cotesting, but I would not want to assume that without further looking into the guidelines and reviewing with a clinical expert. I wonder if proposed modifications to the 'Details' text should be handled in a separate PR? I could add TODOs to this PR, if you think this would be helpful.

cql/ManageRareAbnormality.cql
mickohanlon23: Similar to above, I didn't intend to change any of the details text as a part of this PR, and I think doing so would require further clinical review. But generally, there are other 'Surveillance' `short` recommendations that don't explicitly say "Surveillance" in the `details` text.

cql/ManageSpecialPopulation.cql
mickohanlon23: Should we change to 'Colposcopy and Cytology' which is used elsewhere?

cql/ManageSpecialPopulation.cql
mickohanlon23: This one is tricky because it combines several recommendations, since this is the recommendation for if a patient has a hysterectomy but with no date. They should first be receiving 3 annual tests before doing the 3-year surveillance, so I think adding '3 year' could potentially be misleading if they are still on the annual testing schedule. I would be more inclined to revert this back to 'See Details' if you didn't think 'Surveillance' was appropriate here.

mickohanlon23:
@mrnosal I just responded to your comments here. I didn't make any changes yet - it could be helpful to discuss these in-person to figure out the best way to proceed.

mrnosal (approved):
Mick will open new ticket for addressing detail text changes. Short text is fine for now.

mickohanlon23:
> Mick will open new ticket for addressing detail text changes. Short text is fine for now.

[CCSMCDS-329](https://jira.mitre.org/browse/CCSMCDS-329)

PR #144, opened 2024-09-12
Add recreated pilot patients to test

mrnosal:
This is a set of synthetic test patients recreated from pilot site.


mickohanlon23:
Closing this PR, since we decided not to merge all of these patients into our main branch. We will instead choose a subset of these patients to serve as examples of realistic test patients that we include in our main branch. We will also create additional patient scenarios, since the one's in this set are not representative of all logic paths in our CDS.

PR #145, opened 2024-10-10
Update content type

rdingwell:
The content type defined the FSH and json Library files were incorrectly set to text/cql for elm data.  This PR fixes that issue and updates the packager to include both the elm and cql in the FSH and FHIR json files with correct mime type mappings.


KeeyanGhoreshi:
changes to the resources are mostly superseded by changes made on the cql-libraries branch, but changes to the packager could still be useful.  Unless we aren't planning on merging the changes from cql-libraries into main, I'd say this PR doesn't need to update the fsh.

PR #151, opened 2024-10-29
CCSMCDS-254 Prevent ManagementLibrary from generating erroneous risk table result

mrnosal:
If ManageSpecialPopulation or ManageRareAbnormality generated a recommendation first, the code would still try to show a risk table summary, even though it wouldn't apply to the patient. 

This code checks to see that the recommendation is from ManageCommonAbnormality before producing the RiskTableSummary or the RecommendationFromRiskTable values.


cql/ManagementLibrary.cql
```diff
@@ -16,15 +16,11 @@ library ManagementLibrary version '1.0.0'
 
 using FHIR version '4.0.1'
 
-include CDSConnectCommonsforFHIRv401 version '1.0.0' called C3F
```
mickohanlon23, line 19: I'm curious how you chose which libraries to remove, since there are other included libraries that are also not used in this file?

mickohanlon23 (commented):
Nice thanks for handling this! It's interesting that we had the code in place to check for whether the other recommendations existed, but then commented it out. Merge conflicts will need to be resolved since I took so long to review it. I left one comment about the removed libraries. It's not a huge deal to address here, but if we wanted to remove unused libraries, we may want to consider opening a separate ticket to do so throughout our CQL codebase.

cql/ManagementLibrary.cql
mrnosal: I don't understand the comment. I removed C3F, FHIRHelpers, Common and RiskTableLookup. These were not used in this file. I left Dash, ManageCommonAbnormality, ManageRareAbnormality, ManageSpecialPopulation, Symptomatic and Orders. These libraries are all used in this file.

mickohanlon23: Whoops, I was searching the file by the Library name as opposed to the alias that it's called by. You're right, just the libraries that weren't used in this CQL file were removed.

PR #158, opened 2025-02-19
Add support for CPT-coded Cone treatment

mrnosal:
This addresses the issue of unrecognized treatments causing incorrect recommendations. These treatments were coded with a CPT code instead of a SNOMEDCT code, so were not recognized as a treatment. 
This adds support for this specific CPT code as a direct code in Dashboard.cql
If additional CPT codings are required, then we should consider moving to a new/updated value set.
Adds CPT coded treatment resources in test files looking for cervix excision in patient history.
https://jira.mitre.org/browse/CCSMCDS-326


mickohanlon23 (approved):
Looks good. Let's check with our pilot partner to ensure that this is the primary CPT code they expect to see for a cervical excision. If there are more, we should consider making a value set.

PR #164, opened 2025-03-27
Updated recommendation from errata to include alternative to colposcopy

mrnosal:
This is a request from UMMC:
The errata recommendation for colposcopy should be amended to include an alternative action (1-year follow-up) if the test follows a low-grade result.


cql/ManageCommonAbnormality.cql
```diff
@@ -462,7 +462,8 @@ define ErrataRecommendation:
         short: 'Colposcopy',
         date: Today(),
         details: {
-          'Colposcopy is recommended.'
```
mickohanlon23, line 465: Note to update the L2 document before this is merged.

cql/ManageCommonAbnormality.cql
```diff
@@ -462,7 +462,8 @@ define ErrataRecommendation:
...
+          'Colposcopy is recommended.',
```
mickohanlon23, line 465: Can you point me towards where this evidence come from in the erratum?

cql/ManageCommonAbnormality.cql
```diff
@@ -462,7 +462,8 @@ define ErrataRecommendation:
...
+          'If a patient has a negative HPV test or co-test following a low-grade result for which colposcopy was previously recommended but not performed, repeating an HPV test or cotest in 1 year is acceptable. (Perkins 2019 Erratum)'
```
mickohanlon23, line 466:
```suggestion
          'If a patient has a negative HPV test or cotest following a low-grade result for which colposcopy was previously recommended but not performed, repeating an HPV test or cotest in 1 year is acceptable.'
```
Comments about recommendation text:
1. `co-test` --> `cotest` to align with other recommendations
2. Is there a reason the source document is referenced in this recommendations, when it isn't referenced in most of the other recommendation statements? Should this be added to the "Group" field instead?

cql/ManageCommonAbnormality.cql
```diff
@@ -462,7 +462,8 @@ define ErrataRecommendation:
...
         }
       }
```
mickohanlon23, line 468: Why exactly is this being added here? If a patient's most recent HPV test or cotest were negative, they wouldn't fall into this recommendation, right?

mickohanlon23 (changes requested):

cql/ManageCommonAbnormality.cql
mrnosal: This recommendation for "Colposcopy" is coming from the logic implemented for the Errata recommendations.
(2019 ASCCP Risk-Based Management Consensus Guidelines: Updates through 2023, J Low Genit Tract Dis 2024;28: 3-6)

The pertinent Errata section:

Patient Scenarios Not Initially Addressed in the 2019 Guidelines:
(b) Guideline for cases in which colposcopy was previously recommended but not completed: In cases in which a colposcopy was previously recommended but not completed, the recommendation is for colposcopy if the previous result was high-grade cytology [atypical squamous cells cannot exclude high-grade squamous intraepithelial lesion (ASC-H), atypical glandular cells (AGC), HSIL, or a more severe cytologic interpretation]. If the previous cytology result was not high grade, and the patient undergoes repeat testing with HPV testing or cotesting instead of colposcopy: colposcopy is recommended if the result on repeat testing indicates a second consecutive HPV-positive result and/or persistent cytologic abnormality (ASC-US or a more severe cytologic interpretation); repeat HPV testing or cotesting in 1 year is acceptable if the result on repeat testing is HPV negative or cotest negative.

cql/ManageCommonAbnormality.cql
mickohanlon23: I believe the Perkins 2019 Erratum is different than the 2023 erratum, correct? In that case, it seems confusing to list "Perkins 2019 Erratum" in the recommendation itself but then also list "2023 JLGTD Addendum" in the source section of the recommendation.

cql/ManageCommonAbnormality.cql
mickohanlon23: Section 3.3.3 in the L2 should be updated to include text from https://journals.lww.com/jlgtd/fulltext/2024/01000/2019_asccp_risk_based_management_consensus.2.aspx section 4b:

> If the previous cytology result was not high grade, and the patient undergoes repeat testing with HPV testing or cotesting instead of colposcopy: colposcopy is recommended if the result on repeat testing indicates a second consecutive HPV-positive result and/or persistent cytologic abnormality (ASC-US or a more severe cytologic interpretation); repeat HPV testing or cotesting in 1 year is acceptable if the result on repeat testing is HPV negative or cotest negative.

cql/ManageCommonAbnormality.cql
mickohanlon23: Keeping this comment open, as a reminder that the updated recommendation text should state the following:

> If the previous cytology result was not high grade, and the patient undergoes repeat testing with HPV testing or cotesting instead of colposcopy: colposcopy is recommended if the result on repeat testing indicates a second consecutive HPV-positive result and/or persistent cytologic abnormality (ASC-US or a more severe cytologic interpretation); repeat HPV testing or cotesting in 1 year is acceptable if the result on repeat testing is HPV negative or cotest negative.

cql/ManageCommonAbnormality.cql
mickohanlon23: 1. This will be addressed, as the recommendation text will be updated to the following:

> If the previous cytology result was not high grade, and the patient undergoes repeat testing with HPV testing or cotesting instead of colposcopy: colposcopy is recommended if the result on repeat testing indicates a second consecutive HPV-positive result and/or persistent cytologic abnormality (ASC-US or a more severe cytologic interpretation); repeat HPV testing or cotesting in 1 year is acceptable if the result on repeat testing is HPV negative or cotest negative.

2. The group field should not need to be updated any more, since the text excerpt in our recommendation will be coming directly from the erratum, which is already listed in the group field.

mickohanlon23 (changes requested):
Updated review comments in response to this morning's meeting with pilot clinical advisor.

mrnosal:
Updated 3.2.2.a with additional guideline text. Added test based on UMMC test case for this guideline.

cql/ManageCommonAbnormality.cql
mickohanlon23: We chatted on Slack, and decided that this new text would apply only to 3.2.2.a, and not to 3.2.2.b, since 3.2.2.b deals with patients who have successive HPV+ results, which should result in a colposcopy recommendation.

mickohanlon23 (approved):
Nice! Recommendation was updated based upon our discussion this morning, and a test was added to show that this recommendation is generated for the example scenario we have discussed. Approving, pending the L2 be updated to reflect this change (I think this is a good idea, but open to other thoughts if you believe we shouldn't update the L2 based upon this change). Once a decision is made regarding the L2, you can go ahead and merge this!

PR #171, opened 2025-04-24
update packager and resources

KeeyanGhoreshi:
updates the packager script, gets rid of the library reference in expressions, uses `cql-identifier` instead of `cql` (this will need a compatible change on the encender side), and fixes some warnings and a minor error with the fsh resource generation.


mickohanlon23 (commented):
@KeeyanGhoreshi can you point me towards what FHIR spec update this is intended to comply with? If not sure, it may be helpful to ask @rdingwell to review.

KeeyanGhoreshi:
Not sure which spec, since it's mostly brought over from Rob's PR.

This PR is ready to go, I just tacked on a small change from another related task, since it also updates the packager script. Makes more sense for that to happen in this PR than an entirely separate PR for that change.

mickohanlon23:
> Not sure which spec, since it's mostly brought over from Rob's PR.
> 
> This PR is ready to go, I just tacked on a small change from another related task, since it also updates the packager script. Makes more sense for that to happen in this PR than an entirely separate PR for that change.

So this also completes [CCSMCDS-240](https://jira.mitre.org/browse/CCSMCDS-240) now? Along with [CCSMCDS-237](https://jira.mitre.org/browse/CCSMCDS-237) and [CCSMCDS-239](https://jira.mitre.org/browse/CCSMCDS-239) which it already addressed, right?

Can you reach out to Rob to ask him then?

KeeyanGhoreshi:
Yes, this PR will close the rest of of [224](https://jira.mitre.org/browse/CCSMCDS-224).

The updates include moving in line with the R5 spec by fixing the MIME types for text/cql and application/json+elm.  The update moving from cql -> cql-identifier is not strictly necessary in R5, but cql-identifier is more specific (and used by HAPI).  

Removing the reference from the expression is just a code cleanliness change since it's superfluous for the expressions in our definitions, since Activity/PlanDefinitions will just pull the library from the top level element instead of the expression reference anyway.

fsh-tank/input/fsh/CervicalCancerManagementActivity.fsh
```diff
@@ -21,10 +21,9 @@ Usage: #definition
...
 * dynamicValue[+].path = "orderDetail" // this is the ServiceRequest.orderDetail element
-* dynamicValue[=].expression.language = $EXLANG|4.0.1#text/cql "CQL"
+* dynamicValue[=].expression.language = $EXLANG|4.0.1#text/cql-identifier "CQL"
 * dynamicValue[=].expression.expression = "ServiceRequestOrderDetail"
-* dynamicValue[=].expression.reference = "Library/ManagementLibrary|1.0"
 * dynamicValue[+].path = "occurrenceDateTime" // this is the ServiceRequest.occurrence element
-* dynamicValue[=].expression.language = $EXLANG|4.0.1#text/cql "CQL"
+* dynamicValue[=].expression.language = $EXLANG|4.0.1#text/cql-identifier "CQL"
 * dynamicValue[=].expression.expression = "OrderDate"
 * dynamicValue[=].expression.reference = "Library/ManagementLibrary|1.0"
```
mickohanlon23, line 29: There are several instances where the `expression.reference` is removed from one expression element but not from the secondary one. Is this intentional?

fsh-tank/input/fsh/CervicalCancerScreeningActivity.fsh
```diff
@@ -21,10 +21,9 @@ Usage: #definition
 * intent = $RINTENT#proposal "Proposal"
 * priority = $RPRIOR#routine "Routine"
 * dynamicValue[+].path = "orderDetail" // this is the ServiceRequest.orderDetail element
-* dynamicValue[=].expression.language = $EXLANG|4.0.1#text/cql "CQL"
+* dynamicValue[=].expression.language = $EXLANG|4.0.1#text/cql-identifier "CQL"
 * dynamicValue[=].expression.expression = "ServiceRequestOrderDetail"
-* dynamicValue[=].expression.reference = "Library/ScreeningLibrary|1.0"
```
mickohanlon23, line 26: The Jira ticket says to remove "the reference attributes in the action expression fields in the ActivityDefinition and PlanDefinition resources", however, these ones are in the dynamic value field. Do you know whether or not these should be removed? Whatever we decide, we should keep it consistent between the two dynamic value entries (here one library is removed, but one remains).

util/resources/transactionBundle.json
mickohanlon23, line 1: What is the best way to test that this transaction bundle works?

util/packager.js
```diff
@@ -1,82 +1,129 @@
...
     }
   }
 }
 
 // Loop over all generated FHIR resources and convert them to js files.
 files = readdirSync(fhirPath);
-exportBase = join('dist','fhir');
-if (!existsSync(exportBase)) mkdirSync(exportBase, {recursive: true});
+exportBase = join("dist", "fhir");
+const bundleFilepath = bundlePath + sep + "transactionBundle.json";
```
mickohanlon23, line 98: I wonder if we should also place this transaction bundle with the rest of the exported resources in `dist/`?

mickohanlon23 (changes requested):
Everything compiled correctly. I tested by linking the Dashboard to local versions of ccsm-cds-with-tests and encender, and the Dashboard still works as intended.

I had a few suggestions and questions that I left in comments.

util/resources/transactionBundle.json
mickohanlon23: We discussed today, and decided @KeeyanGhoreshi will document steps for posting resources to the HAPI server via a transaction bundle.

util/packager.js
```diff
@@ -1,82 +1,129 @@
...
+        lines = lines.slice(0, descriptionIndex + 1);
+      }
+
+      const updatedContents = `${lines.join('\n')}
+* content[cqlContent].contentType = urn:ietf:bcp:13#text/cql "text/cql"
+* content[cqlContent].data = "${cql.toString("base64")}"
+* content[1].contentType = urn:ietf:bcp:13#application/elm+json "application/elm+json"
+* content[1].data = "${contentBase64}"
+
+`;
```
mickohanlon23, line 62: If I'm understanding correctly, this adds a dependency for the content field to be immediately after the description field, which may not always be the case. Should we restore the old behavior where we explicitly replace the content.data elements?

mickohanlon23 (changes requested):
I just reviewed the packager.js and transactionBundle.json updates more closely. I left one comment. It would still be nice to:
1.  See a list of steps for testing the transactionBundle against a HAPI Server.
2.  Remove the extra linting in the packager.js file from this PR, if possible (I asked Rob if there is an easy way to do this).

util/packager.js
KeeyanGhoreshi: it's either after the description field or tacked on to the end.  I'm not sure why this change was made, you'd have to ask Rob.  Though it shouldn't matter, the FSH files are local.  If a file is missing the description or has the description in the wrong place we can just fix it.

fsh-tank/input/fsh/CervicalCancerScreeningActivity.fsh
KeeyanGhoreshi: yeah I think they all should be removed dynamic or not.  They all should pull from the library element.  I think the only exception would be if there was more than 1 library in the list, in which case they'd have to specify.  But I think that's only an issue with the flat plandefs.

util/packager.js
KeeyanGhoreshi: I don't think so, it's more for convenience than it is a real bundle resource.

PR #173, opened 2025-05-21
Adjust processing HPV results for risk table lookups

mrnosal:
Addressing [CCSMCDS-373](https://jira.mitre.org/browse/CCSMCDS-373)

AutogeneratedTableLookup.cql contains the functions that try to match patient history with entries in the autogenerated risk tables (AutogeneratedRiskTables.cql).

The issue encountered is that HPV results can be interpreted as four results: HPV-negative, HPV-positive, HPV16+, HPV16-,18+. Some of the lookup functions are looking only for a result that is "HPV-positive". If the result is "HPV16+", those functions may fail to see the result as a positive HPV result. 

The current lookup functions for Table 3 make use of the AnyHpv() function to treat HPV16+ and HPV16-,18+ as "HPV-positive" results. 

These changes expand on the use of the AnyHpv() function, and expand the test cases to include HPV16+ and HPV18+ results, not just non-typed hrHPV "HPV-positive".

For Table 1 lookups, there are individual rows that match HPV16+ and HPV18+. The Table 1 history does not look for HPV-positive results. Therefore, no changes are required for Table 1 lookups.

Table 2 lookups rely on the HPV history and history-prev-2, so these values are sent through AnyHpv() to recognize HPV16+ and HPV18+. Current HPV row values do have entries for HPV16+ and HPV18+

Table 3 already uses AnyHpv, so no additional changes needed.

Table 4 relies on "preColposcopyResult" so it will need to use AnyHpv()

Table 5 relies on current HPV value, so these get wrapped in AnyHpv(). LongTermSurveillance is only looking at HPV-negative results, so no change needed there. 

Test cases are expanded by adding HPV16+ and HPV18+ results to the resources.yml files that the tests are iterating over. 

The RarelyScreened Post-Treatment test cases, I believe, were erroneously giving recommendations from Table 1 instead of Table 5 because of this particular issue.

### DIFF
--- a/cql/AutogeneratedTableLookup.cql
+++ b/cql/AutogeneratedTableLookup.cql
@@ -184,6 +184,9 @@ define function GetSurveillanceManagement(prev2HpvVal String, prev2PapVal String
     }
 
 define function GetRowTable2(prev2HpvVal String, prev2PapVal String, prevHpvVal String, prevPapVal String, currHpvVal String, currPapVal String):
+  _GetRowTable2(AnyHpv(prev2HpvVal), prev2PapVal, AnyHpv(prevHpvVal), prevPapVal, currHpvVal, currPapVal)
+  
+define function _GetRowTable2(prev2HpvVal String, prev2PapVal String, prevHpvVal String, prevPapVal String, currHpvVal String, currPapVal String):
   Coalesce(
     Coalesce( // Case 1: Current cytology is fully specified
       SurveillanceTable T
@@ -474,6 +477,9 @@ define function GetPostColposcopyManagement(preBiopsyHpvVal String, preBiopsyPap
     }
 
 define function GetRowTable4(preBiopsyHpvVal String, preBiopsyPapVal String, biopsyVal String, prev2HpvVal String, prev2PapVal String, prevHpvVal String, prevPapVal String, currHpvVal String, currPapVal String):
+  _GetRowTable4(AnyHpv(preBiopsyHpvVal), preBiopsyPapVal, biopsyVal, prev2HpvVal, prev2PapVal, prevHpvVal, prevPapVal, AnyHpv(currHpvVal), currPapVal)
+
+define function _GetRowTable4(preBiopsyHpvVal String, preBiopsyPapVal String, biopsyVal String, prev2HpvVal String, prev2PapVal String, prevHpvVal String, prevPapVal String, currHpvVal String, currPapVal String):
   if biopsyVal = '<CIN1' or biopsyVal = 'CIN1' then
     Coalesce(
       Coalesce( // Case 1: Explicit current cytology
@@ -566,9 +572,9 @@ define function GetRowTable4(preBiopsyHpvVal String, preBiopsyPapVal String, bio
 
 define function GetAdjacentRowsTable4(preBiopsyHpvVal String, preBiopsyPapVal String, biopsyVal String, prev2HpvVal String, prev2PapVal String, prevHpvVal String, prevPapVal String, currHpvVal String, currPapVal String):
   PostColposcopyTable T
-    where T.preColposcopyResult = HighOrLowGradeResults(preBiopsyHpvVal, preBiopsyPapVal)
+    where T.preColposcopyResult = HighOrLowGradeResults(AnyHpv(preBiopsyHpvVal), preBiopsyPapVal)
       and T.postColposcopyPastHistory = PostColposcopyPastHistory(prev2HpvVal, prev2PapVal, prevHpvVal, prevPapVal)
-      and T.hpv = currHpvVal
+      and T.hpv = AnyHpv(currHpvVal)
       and T.pap != currPapVal
       return {
         action: T.management,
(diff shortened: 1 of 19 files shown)


test/RarelyScreened/cases/HsilHighRiskHpv16PositiveRecentPrecancerTreatment_NotRS.yml
```diff
@@ -28,11 +28,11 @@ data:
 
 results:
```
mickohanlon23, line 29: For these RarelyScreened test cases, now that they are receiving recommendations from Table 5, I wonder if we should give them a biopsy before their treatment, so that we are not relying on CollateManagementData.cql's [ManageTreatmentWithoutPreviousBiopsy](https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/blob/main/cql/CollateManagementData.cql#L34) parameter to be `true`.

cql/AutogeneratedTableLookup.cql
```diff
@@ -184,6 +184,9 @@ define function GetSurveillanceManagement(prev2HpvVal String, prev2PapVal String
     }
 
 define function GetRowTable2(prev2HpvVal String, prev2PapVal String, prevHpvVal String, prevPapVal String, currHpvVal String, currPapVal String):
+  _GetRowTable2(AnyHpv(prev2HpvVal), prev2PapVal, AnyHpv(prevHpvVal), prevPapVal, currHpvVal, currPapVal)
```
mickohanlon23, line 187: - Does `GetAdjacentRowsTable2` function also need `AnyHpv()` call?
- Does `GetAdjacentRowsTable5` function also need `AnyHpv()` call?

mickohanlon23 (commented):
Good find! Thanks for adding HPV16 and HPV18 to the relevant "HPV positive" scenarios in the test cases. The test case updates make sense to me. As you pointed out last week, this likely addresses some of what was intended by [CCSMCDS-293](https://jira.mitre.org/browse/CCSMCDS-293).

It looks like there are merge conflicts now, probably because I merged the update to all the recommendation IDs. Additionally, wasn't there a scenario we found that was left out when we were reviewing this briefly together last week? 

Otherwise this looks good to me, pending resolution of two comments I added.

test/RarelyScreened/cases/HsilHighRiskHpv16PositiveRecentPrecancerTreatment_NotRS.yml
mrnosal: Added biopsy (CIN2/CIN3) to these cases.

cql/AutogeneratedTableLookup.cql
mrnosal: Added appropriate AnyHpv() calls to GetAdjacentRowsTable2. GetAdjacentRowsTable5 doesn't need AnyHpv() because the LongTermSurveillance function is only looking for associated HPV-negative values.

mrnosal: I also added test resources to the ManagementTable2 tests to iterate over HPV+, HPV16+ and HPV16-,18+ results in the test case histories to exercise finding the adjacent rows in table 2.

mickohanlon23: The added `AnyHPV()` calls to `GetAdjacentRowsTable2` look good.

mickohanlon23: I see why we currently wouldn't need the `AnyHPV()` call for Table 5. But do you know why the GetRowTable5 function only looks for the scenarios where the HPV input is one of the multiple-negative-test scenarios? This is technically not directly related to this PR, so shouldn't hold this up from being merged, but it could be worth some investigation in the future.

mickohanlon23: I opened [CCSMCDS-400](https://jira.mitre.org/browse/CCSMCDS-400) to prompt an investigation about the GetAdjacentRowsTableX logic. @mrnosal if you have already done this investigation, feel free to close the Jira ticket, or move it into the current sprint as you see fit.

cql/AutogeneratedTableLookup.cql
```diff
@@ -184,11 +184,14 @@ define function GetSurveillanceManagement(prev2HpvVal String, prev2PapVal String
...
+
+define function _GetRowTable2(prev2HpvVal String, prev2PapVal String, prevHpvVal String, prevPapVal String, currHpvVal String, currPapVal String):
   Coalesce(
     Coalesce( // Case 1: Current cytology is fully specified
       SurveillanceTable T
-      where T.history_prev_2 = ScreeningText(prev2HpvVal, prev2PapVal, '/', 'none')
-        and T.history = historyLeadingSpace(ScreeningText(prevHpvVal, prevPapVal, '/', 'none'))
+      where T.history_prev_2 = ScreeningText(AnyHpv(prev2HpvVal), prev2PapVal, '/', 'none')
```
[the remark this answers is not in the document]
mickohanlon23: @mrnosal Why were the `AnyHpv()` calls also removed from `GetAdjacentRowsTable2()`? Aren't they still needed there, since `GetAdjacentRowsTableX` functions don't leverage the converted values from the wrapper function?

cql/AutogeneratedTableLookup.cql
mrnosal: I see - this will require additional investigation. "AdjacentRows" is our concept, so the behavior is determined by us.

PR #174, opened 2025-05-29
Document CQL parameters used by CCSM CDS

mrnosal:
Create a list of all our customizable CQL parameters, and what each of them does:
~1 sentence description of what parameter is used for
Mention default value (if parameter isn't set)


mickohanlon23 (approved):
This is extremely thorough, thank you!

PR #176, opened 2025-07-01
Ccsmcds 383 log additional data

mrnosal:
Log additional data from the CQL. Some of these are required for the analytics (e.g. DateOfFirstBiopsyAfterMostRecentHpvReport) and others should be logged because the CQL logic behind them is difficult to replicate in the analytics stage. The log files are not large, so adding these isn't a file size concern.


mickohanlon23 (commented):
Looks good to me. I tested by installing this as a locally dependency for the dashboard with the logger turned on, and I could see the newly logged data elements coming through.

Have you done any performance testing with this update? Just want to make sure it is not significantly slowing down processing time.

mrnosal:
> Looks good to me. I tested by installing this as a locally dependency for the dashboard with the logger turned on, and I could see the newly logged data elements coming through.
> 
> Have you done any performance testing with this update? Just want to make sure it is not significantly slowing down processing time.

I've noticed no noticable performance difference with these changes.

mickohanlon23 (approved):
Approving. PR looks good, and @mrnosal just confirmed that he has seen no performance changes using this in his testing. Let's keep an eye out for for any performance differences as we deploy this with our pilot partner.

PR #177, opened 2025-07-03
Coerce previous HPV results for Table 2 adjacent row lookup

mrnosal:
Needed to restore the AnyHpv() call to GetAdjacentRowsTable2 - previous HPV results needed to be coerced to "HPV-positive" to find adjacent rows if the result was HPV16+ or HPV16-,18+

Verify via inspection of ManagementTable2 result .json files and inspect the Table 2 Adjacent Rows array for cases where the previous HPV results were 16 or 18. Without the fix, they are empty arrays.


mickohanlon23 (approved):
This addresses the following comment: https://github.com/ccsm-cds-tools/ccsm-cds-with-tests/pull/173#discussion_r2180742080.

Thanks!

PR #179, opened 2025-07-03
Return referring hpv and cyto values from table3 lookups

mrnosal:
Table3 lookups don't return the referring HPV and Cytology result values. Add these values to be consistent with the other table lookup returns.

### DIFF
--- a/cql/AutogeneratedTableLookup.cql
+++ b/cql/AutogeneratedTableLookup.cql
@@ -327,8 +327,8 @@ define function GetRowTable3(preBiopsyHpvVal String, preBiopsyPapVal String, bio
             biopsy: T.biopsy,
             preColposcopyResult: '',
             postColposcopyPastHistory: '',
-            hpv: '',
-            cyto: '',
+            hpv: preBiopsyHpvVal,
+            cyto: preBiopsyPapVal,
             riskNow: convert T.value.immediate to Decimal,
             risk5yr: convert T.value.five_year to Decimal
           }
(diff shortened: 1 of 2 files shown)


mrnosal:
> Is this a change that's needed for the Analytics output?
Yes - when computing metrics on recommendations from the risk tables.
> 
> The other Tables are only returning values that come from the tables themselves in the `return` of the `GetRowTableX` functions. I believe the `referral` element was created specifically to return the HPV and pap information from Table 3, since Table 3 is the one table that doesn't have a current "pap" or current "hpv" value, but combines them into 1 "Referral Screen Result" column.

The whole AutogeneratedTableLookup part involves these awkward string manipulation functions just to get values that match what was in the published tables. Reverse-engineering these for use in analytics is doubly awkward.

> I might suggest we keep this as-is so that the return values of these functions are solely focused on returning the data that comes from the table. But if it is necessary for analytics, we could reconsider, or even make a change so that the preBiopsyHpvVal and preBiopsyPapVal can be discovered elsewhere.

There is nothing that currently relies on or takes advantage of the fact that the returned value matches the autogenerated risk table value. CCSMCDS-401 suggests a different approach which would remove these complex string combinations entirely. Having the hpv and pap values returned here would at least make it similar processing results from other risk tables.

mickohanlon23 (approved):
We discussed this at a recent Dev Meeting. This change is helpful to power the analytics library. And it only changes return values that we are not currently doing anything with. As was suggested, we can consider modifying how these value are returned as a part of [CCSMCDS-401](https://jira.mitre.org/browse/CCSMCDS-401).

PR #181, opened 2025-08-25
Add PertinentImmunizationsSummary to analytics

mrnosal:
Pilot has asked for analytics based off HPV immunization dates. This is not currently used in the CDS logic, but it is captured and displayed in the Dashboard if available. 
Adds PertinentImmunizationsSummary to the DashboardLibrary analytics object.


mickohanlon23 (approved):
This just adds `Dash.PertinentImmunizationsSummary` to the AnalyticLibrary, so that it can be used in Analytics. @mrnosal go ahead and merge when you are ready.

